PR #3, opened 2022-04-04
Support loading uncertainties in NXdata

SimonHeybrock:
Note that this is also used for NXlog, NXmonitor, etc.

### DIFF
--- a/tests/nxdata_test.py
+++ b/tests/nxdata_test.py
@@ -217,3 +217,46 @@ def test_create_field_from_variable(nxroot, unit):
 def test_create_class(nxroot, nx_class):
     group = nxroot.create_class('group', nx_class)
     assert group.nx_class == nx_class
+
+
+@pytest.mark.parametrize("errors_suffix", ['_error', '_errors'])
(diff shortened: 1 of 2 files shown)


tests/nxdata_test.py
```diff
@@ -217,3 +217,46 @@ def test_create_field_from_variable(nxroot, unit):
...
+    data = nxroot.create_class('data1', NX_class.NXdata)
+    data.attrs['axes'] = da.dims
+    data.attrs['signal'] = 'signal'
+    data.create_field('signal', da.data)
+    data.create_field(f'xx{errors_suffix}', da.coords[f'xx{errors_suffix}'])
+    assert sc.identical(data[...], da)
+
+
+@pytest.mark.parametrize("errors_suffix", ['_error', '_errors'])
+def test_uncertainties_of_coords_are_loaded(nxroot, errors_suffix):
```
nvaytet, line 237: Not sure I followed everything but is this only for loading uncertainties of coords, not uncertainties of data?

SimonHeybrock: Loading the latter has already been supported for a while.

nvaytet (approved):

PR #4, opened 2022-04-04
Basics docs

SimonHeybrock:
A lot remains to be added (in particular also docstrings), but this is a start.

### DIFF
--- a/docs/getting-started/quick-start-guide.ipynb
+++ b/docs/getting-started/quick-start-guide.ipynb
@@ -306,8 +306,7 @@
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.8.12"
+   "pygments_lexer": "ipython3"
   }
  },
  "nbformat": 4,
--- a/docs/getting-started/quick-start-guide.ipynb
+++ b/docs/getting-started/quick-start-guide.ipynb
@@ -29,7 +29,7 @@
     "  - Attributes that identify which fields contained in the group hold signal values, and which hold axis labels.\n",
     "  \n",
     "In the following we use a file from the [POWGEN](https://neutrons.ornl.gov/powgen) instrument at SNS.\n",
-    "It is bundled with scippnexus and will be downloaded automatically using [pooch](https://pypi.org/project/pooch/) if it is not cahced already:"
+    "It is bundled with scippnexus and will be downloaded automatically using [pooch](https://pypi.org/project/pooch/) if it is not cached already:"
    ]
   },
   {
@@ -187,11 +187,11 @@
     "  - Units of the fields.\n",
     "  - Uncertainties of the fields (here for `'average_value'`).\n",
     "  \n",
-    "This structure is compatible to a `scipp.DataArray` and is returned as such.\n",
+    "This structure is compatible with a `scipp.DataArray` and is returned as such.\n",
     "\n",
-    "We may also load an individual fields instead of an entire group.\n",
+    "We may also load an individual field instead of an entire group.\n",
     "A field corresponds to a `scipp.Variable`, i.e., similar to how h5py represents datasets as NumPy arrays but with an added unit and dimension labels (if applicable).\n",
-    "For example, we may load only value dataset:"
+    "For example, we may load only the `'value'` dataset:"
    ]
   },
   {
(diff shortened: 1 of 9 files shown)


docs/getting-started/quick-start-guide.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "\n",
+    "- Definitions for attributes for datasets, in particular a `units` attribute.\n",
+    "  In NeXus, datasets are referred to as *field*.\n",
+    "- Definitions for attributes and structure of groups.\n",
+    "  This includes:\n",
+    "  - An `NX_class` attribute, identifying a group as an instance of a particular NeXus class such as [NXdata](https://manual.nexusformat.org/classes/base_classes/NXdata.html) or [NXlog](https://manual.nexusformat.org/classes/base_classes/NXlog.html).\n",
+    "  - Attributes that identify which fields contained in the group hold signal values, and which hold axis labels.\n",
+    "  \n",
+    "In the following we use a file from the [POWGEN](https://neutrons.ornl.gov/powgen) instrument at SNS.\n",
+    "It is bundled with scippnexus and will be downloaded automatically using [pooch](https://pypi.org/project/pooch/) if it is not cahced already:"
```
nvaytet, line 32:
```suggestion
    "It is bundled with scippnexus and will be downloaded automatically using [pooch](https://pypi.org/project/pooch/) if it is not cached already:"
```

docs/getting-started/quick-start-guide.ipynb
```diff
@@ -0,0 +1,314 @@
...
+   "source": [
+    "Above, scippnexus automatically dealt with:\n",
+    "\n",
+    "- Loading the data field (signal value dataset and its `'units'` attribute).\n",
+    "- Identifying the dimension labels (here: `'time'`).\n",
+    "- Other fields in the group were loaded as coordinates, including:\n",
+    "  - Units of the fields.\n",
+    "  - Uncertainties of the fields (here for `'average_value'`).\n",
+    "  \n",
+    "This structure is compatible to a `scipp.DataArray` and is returned as such.\n",
```
nvaytet, line 190:
```suggestion
    "This structure is compatible with a `scipp.DataArray` and is returned as such.\n",
```

docs/getting-started/quick-start-guide.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "\n",
+    "We may also load an individual fields instead of an entire group.\n",
```
nvaytet, line 192:
```suggestion
    "We may also load an individual field instead of an entire group.\n",
```
(or `load individual fields`, whichever you prefer)

docs/getting-started/quick-start-guide.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "A field corresponds to a `scipp.Variable`, i.e., similar to how h5py represents datasets as NumPy arrays but with an added unit and dimension labels (if applicable).\n",
+    "For example, we may load only value dataset:"
```
nvaytet, line 194:
```suggestion
    "For example, we may load only the `'value'` dataset:"
```

docs/getting-started/quick-start-guide.ipynb
```diff
@@ -0,0 +1,314 @@
...
+   "source": [
+    "proton_charge['value'].attrs['units']"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d75fa4b1-8490-49c6-809b-fdc930960886",
+   "metadata": {},
+   "source": [
+    "A subset of the group (and its datasets) can loaded loading by selecting only a slice.\n",
```
nvaytet, line 230:
```suggestion
    "A subset of the group (and its datasets) can be loaded by selecting only a slice.\n",
```

docs/user-guide/classes.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Typing
+------
+
+.. autosummary::
+   :toctree: ../generated/classes
+   :template: scipp-class-template.rst
+   :recursive:
+
+   typing.Dataset
```
nvaytet, line 36: Is this going to be very confusing with the scipp `Dataset` ?

nvaytet (changes requested):

docs/user-guide/classes.rst
SimonHeybrock: Yes, but we can't do anything about this. Unless we want to rename `scipp.Dataset`?

SimonHeybrock: Well, we could maybe add a prefix, `H5Dataset`?

SimonHeybrock: I will do that.

nvaytet (approved):

PR #5, opened 2022-04-04
Add basic docs on supported Nexus classes

SimonHeybrock:
Edit: I think the long list of redundant sections (for classes that simply group other stuff) is not so useful. I think I will switch this to a table instead, linking to pages generated from docstrings, and a a couple of dedicated sections with useful information (as contained in the current branch, i.e., for NXdata, NXdetector, ...):

image

### DIFF
--- a/docs/user-guide/nexus-classes.ipynb
+++ b/docs/user-guide/nexus-classes.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "444ce621-fecd-4a30-ba53-15bb9c87bd96",
+   "id": "04343de9-cb0d-44ae-a1a8-a107da63a8ff",
    "metadata": {
     "tags": []
    },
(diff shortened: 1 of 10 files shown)


SimonHeybrock:
@nvaytet I pushed an update. This is ready for review now.

docs/user-guide/nexus-classes.ipynb
```diff
@@ -0,0 +1,315 @@
...
+    "import scippnexus as snx\n",
+    "f = snx.File(filename)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f59eda25-466c-4437-a7b4-825cb036232c",
+   "metadata": {},
+   "source": [
+    "NeXus class | read as | comment | NeXus specification\n",
```
nvaytet, line 43: The table is very nice, but i'm a little confused as to why it shows up after the code cell above.
Wouldn't it make more sense to put the table before L22?

SimonHeybrock:
@nvaytet I think I addressed your comments.

nvaytet (approved):

PR #7, opened 2022-04-05
Make squeezing mechanism consistent and generic

SimonHeybrock:
Previously we squeezed only coords added to a data array. When accessing fields directly they returned variables of different dims. Furthermore this avoids the need for squeezing downstream in a number of cases.

### DIFF
--- a/tests/nexus_test.py
+++ b/tests/nexus_test.py
@@ -146,7 +146,7 @@ def test_field_properties(nxroot):
     field = nxroot['entry/events_0/event_time_offset']
     assert field.dtype == 'int64'
     assert field.name == '/entry/events_0/event_time_offset'
-    assert field.shape == (6, )
+    assert field.shape == [6]
     assert field.unit == sc.Unit('ns')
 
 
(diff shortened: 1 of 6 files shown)


tests/nexus_test.py
```diff
@@ -146,7 +146,7 @@ def test_field_properties(nxroot):
     field = nxroot['entry/events_0/event_time_offset']
     assert field.dtype == 'int64'
     assert field.name == '/entry/events_0/event_time_offset'
-    assert field.shape == (6, )
+    assert field.shape == [6]
```
nvaytet, line 149: In numpy, shape is a tuple, not a list. We have it as a list in scipp. Should we change it to a tuple?

nvaytet (approved):
A small comment unrelated to this PR.
Other than that, looks good.

tests/nexus_test.py
SimonHeybrock: Good point, h5py also uses a tuple...

SimonHeybrock: https://github.com/scipp/scipp/issues/2521

PR #8, opened 2022-04-05
Generalize datetime handling

SimonHeybrock:
Previously only a datetime offset attribute for `NXlog['time']` and `event_time_zero` in `NXevent_data` was loaded. This PR removes this special-case handling and instead attempts to do so for *all* fields. Rationale is as follows:

- HDF5 does not support storing datetime64.
- According to private communication with Tobias Richter (on the Nexus committee), interpreting start/offset attributes *is* standard (or in the spirit if the standard), even though this is not documented anywhere.

Related changes:

- Try to preserve precision of date in file instead of converting everything to nanoseconds.
- Work around numpy timezone deprecation warnings.

### DIFF
--- a/src/scippnexus/_common.py
+++ b/src/scippnexus/_common.py
@@ -31,13 +31,22 @@ def convert_time_to_datetime64(
             time series data and the unit of the raw_times Variable. If
             not provided, defaults to 1 (a no-op scaling factor).
     """
-    raw_times_ns = sc.to_unit(raw_times, sc.units.ns, copy=False)
+    if (raw_times.dtype
+            in (sc.DType.float64, sc.DType.float32)) or scaling_factor is not None:
+        unit = sc.units.ns
+    else:
+        # determine more precise unit
+        ratio = sc.scalar(1.0, unit=start.unit) / sc.scalar(
+            1.0, unit=raw_times.unit).to(unit=start.unit)
+        unit = start.unit if ratio.value < 1.0 else raw_times.unit
+
+    raw_times = raw_times.to(unit=unit, copy=False)
 
     if scaling_factor is None:
-        times = raw_times_ns.astype(sc.DType.int64, copy=False)
+        times = raw_times.astype(sc.DType.int64, copy=False)
     else:
         _scale = sc.scalar(value=scaling_factor)
-        times = (raw_times_ns * _scale).astype(sc.DType.int64, copy=False)
+        times = (raw_times * _scale).astype(sc.DType.int64, copy=False)
     return start + times
 
 
--- a/src/scippnexus/_common.py
+++ b/src/scippnexus/_common.py
@@ -40,14 +40,12 @@ def convert_time_to_datetime64(
             1.0, unit=raw_times.unit).to(unit=start.unit)
         unit = start.unit if ratio.value < 1.0 else raw_times.unit
 
-    raw_times = raw_times.to(unit=unit, copy=False)
-
     if scaling_factor is None:
-        times = raw_times.astype(sc.DType.int64, copy=False)
+        times = raw_times
     else:
-        _scale = sc.scalar(value=scaling_factor)
-        times = (raw_times * _scale).astype(sc.DType.int64, copy=False)
-    return start.to(unit=unit, copy=False) + times
+        times = raw_times * sc.scalar(value=scaling_factor)
+    return start.to(unit=unit, copy=False) + times.to(
+        dtype=sc.DType.int64, unit=unit, copy=False)
 
 
 def _to_canonical_select(dims: List[str], select: ScippIndex) -> ScippIndex:
--- a/src/scippnexus/_common.py
+++ b/src/scippnexus/_common.py
@@ -9,7 +9,6 @@ from .typing import ScippIndex
 
 def convert_time_to_datetime64(
         raw_times: sc.Variable,
-        group_path: str,
         start: str = None,
         scaling_factor: Union[float, np.float_] = None) -> sc.Variable:
     """
@@ -25,8 +24,6 @@ def convert_time_to_datetime64(
 
     Args:
         raw_times: The raw time data from a nexus file.
-        group_path: The path within the nexus file to the log being read.
-            Used to generate warnings if loading the log fails.
         start: Optional, the start time of the log in an ISO8601
             string. If not provided, defaults to the beginning of the
             unix epoch (1970-01-01T00:00:00).
@@ -34,31 +31,14 @@ def convert_time_to_datetime64(
             time series data and the unit of the raw_times Variable. If
             not provided, defaults to 1 (a no-op scaling factor).
     """
-    try:
-        raw_times_ns = sc.to_unit(raw_times, sc.units.ns, copy=False)
-    except sc.UnitError:
-        raise sc.UnitError(
-            f"The units of time in the entry at "
-            f"'{group_path}/time{{units}}' must be convertible to seconds, "
-            f"but this cannot be done for '{raw_times.unit}'. Skipping "
-            f"loading group at '{group_path}'.")
-
-    try:
-        _start_ts = sc.scalar(value=np.datetime64(start or "1970-01-01T00:00:00"),
-                              unit=sc.units.ns,
-                              dtype=sc.DType.datetime64)
-    except ValueError:
-        raise ValueError(
-            f"The date string '{start}' in the entry at "
-            f"'{group_path}/time@start' failed to parse as an ISO8601 date. "
-            f"Skipping loading group at '{group_path}'")
+    raw_times_ns = sc.to_unit(raw_times, sc.units.ns, copy=False)
 
     if scaling_factor is None:
         times = raw_times_ns.astype(sc.DType.int64, copy=False)
     else:
         _scale = sc.scalar(value=scaling_factor)
         times = (raw_times_ns * _scale).astype(sc.DType.int64, copy=False)
-    return _start_ts + times
+    return start + times
 
 
 def _to_canonical_select(dims: List[str], select: ScippIndex) -> ScippIndex:
--- a/src/scippnexus/_common.py
+++ b/src/scippnexus/_common.py
@@ -47,7 +47,7 @@ def convert_time_to_datetime64(
     else:
         _scale = sc.scalar(value=scaling_factor)
         times = (raw_times * _scale).astype(sc.DType.int64, copy=False)
-    return start + times
+    return start.to(unit=unit, copy=False) + times
 
 
 def _to_canonical_select(dims: List[str], select: ScippIndex) -> ScippIndex:
--- a/src/scippnexus/docs/our-interpretation-of-the-nexus-format.md
+++ b/src/scippnexus/docs/our-interpretation-of-the-nexus-format.md
@@ -47,6 +47,16 @@ More concretely this means that, e.g., for loading an `NXdetector` from a NH, th
 
 If the above yields no more than one item, the group can be loaded.
 
+## Datetime fields
+
+HDF5 does not support storing date and time information such as `np.datetime64`.
+`NXlog` and `NXevent_data` specify specific attributes for fields that have to be interpreted as date and time, in particular [NXlog/time@start](https://manual.nexusformat.org/classes/base_classes/NXlog.html#nxlog-time-start-attribute) and [NXevent_data/event_time_offset@offset](https://manual.nexusformat.org/classes/base_classes/NXevent_data.html#nxevent-data-event-time-offset-field).
+No *general* definition or intention is documented in the NF, but according to TR this is nevertheless standard.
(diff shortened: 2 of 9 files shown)


tests/nexus_test.py
```diff
@@ -212,6 +212,54 @@ def test_field_of_extended_ascii_in_ascii_encoded_dataset_is_loaded_correctly():
...
+                     unit='ns',
+                     values=['1970-01-01T00:00:00', '1970-01-01T00:00:01']))
+
+
+@pytest.mark.parametrize(
+    'timezone', ['Z', '+04', '+00', '-02', '+1130', '-0930', '+11:30', '-09:30'])
+def test_timezone_information_in_datetime_attribute_is_dropped(nxroot, timezone):
+    nxroot['mytime'] = sc.arange('ignored', 2, unit='s')
+    nxroot['mytime'].attrs['start_time'] = f'1970-01-01T00:00:00{timezone}'
+    print(
```
nvaytet, line 251: remove `print`?

src/scippnexus/docs/our-interpretation-of-the-nexus-format.md
```diff
@@ -47,6 +47,16 @@ More concretely this means that, e.g., for loading an `NXdetector` from a NH, th
...
 
+## Datetime fields
+
+HDF5 does not support storing date and time information such as `np.datetime64`.
+`NXlog` and `NXevent_data` specify specific attributes for fields that have to be interpreted as date and time, in particular [NXlog/time@start](https://manual.nexusformat.org/classes/base_classes/NXlog.html#nxlog-time-start-attribute) and [NXevent_data/event_time_offset@offset](https://manual.nexusformat.org/classes/base_classes/NXevent_data.html#nxevent-data-event-time-offset-field).
+No *general* definition or intention is documented in the NF, but according to TR this is nevertheless standard.
+Due to the attribute naming mismatch in the two cases where it *is* specified we need to assume that naming is arbitrary.
+Therefore, we search *all* attributes of a field for a date and time offset, provided that the field's unit is a time unit.
+It is unclear what should be done in the case of multiple matches.
+As of April 2022 we ignore the offset in this case, since guessing which one to use based on the attribute name does not seem desirable.
```
nvaytet, line 58: Do you have a test for this? Not sure I saw one below.

nvaytet (approved):

src/scippnexus/_common.py
```diff
@@ -25,40 +24,30 @@ def convert_time_to_datetime64(
...
 
     if scaling_factor is None:
-        times = raw_times_ns.astype(sc.DType.int64, copy=False)
+        times = raw_times.astype(sc.DType.int64, copy=False)
     else:
         _scale = sc.scalar(value=scaling_factor)
-        times = (raw_times_ns * _scale).astype(sc.DType.int64, copy=False)
-    return _start_ts + times
+        times = (raw_times * _scale).astype(sc.DType.int64, copy=False)
+    return start.to(unit=unit, copy=False) + times
```
jl-wynen, line 50: Would it make sense to combine the unit and dtype conversions into a single call?
```.py
return start.to(unit=unit, copy=False) + times.to(unit=unit, dtype=sc.DType.int64, copy=False)
```

src/scippnexus/docs/our-interpretation-of-the-nexus-format.md
```diff
@@ -47,6 +47,16 @@ More concretely this means that, e.g., for loading an `NXdetector` from a NH, th
 
 If the above yields no more than one item, the group can be loaded.
 
+## Datetime fields
+
+HDF5 does not support storing date and time information such as `np.datetime64`.
+`NXlog` and `NXevent_data` specify specific attributes for fields that have to be interpreted as date and time, in particular [NXlog/time@start](https://manual.nexusformat.org/classes/base_classes/NXlog.html#nxlog-time-start-attribute) and [NXevent_data/event_time_offset@offset](https://manual.nexusformat.org/classes/base_classes/NXevent_data.html#nxevent-data-event-time-offset-field).
+No *general* definition or intention is documented in the NF, but according to TR this is nevertheless standard.
```
jl-wynen, line 54:
```suggestion
No *general* definition or intention is not documented in the NF, but according to TR this is nevertheless standard.
```
?

src/scippnexus/docs/our-interpretation-of-the-nexus-format.md
jl-wynen, line 58: I cannot follow. What attributes have a mismatch?
What 'offset'? Is this talking about time zones or event_time_offset?

src/scippnexus/nxobject.py
```diff
@@ -84,21 +86,53 @@ def __getitem__(self, name: str) -> Any:
...
+        try:
+            # NumPy and scipp cannot handle timezone information. We therefore strip it,
+            # i.e., interpret time as local time. If time is given in UTC this will lead
+            # to misleading results since we have no information about the actual time
+            # zone.
+            # Would like to use dateutil, but with Python's datetime we do not get
+            # nanosecond precision.
+            if 'T' in obj:
+                date, time = obj.split('T')
+                time = re.split(f'Z|{re.escape("+")}|-', time)[0]
```
jl-wynen, line 119:
```suggestion
                time = re.split(r'Z|\+|-', time)[0]
```

src/scippnexus/nxobject.py
```diff
@@ -137,6 +171,18 @@ def __getitem__(self, select) -> sc.Variable:
             self._dataset.read_direct(variable.values, source_sel=index)
         else:
             variable.values = self._dataset[index]
+        if self._is_time or _is_time(variable):
+            starts = []
+            for name in self.attrs:
+                if (dt := _as_datetime(self.attrs[name])) is not None:
+                    starts.append(dt)
```
jl-wynen, line 178: Does this mean that all string attrs that contain 'T' and can be parsed by numpy are treated as datetimes?

src/scippnexus/nxobject.py
```diff
@@ -137,6 +171,18 @@ def __getitem__(self, select) -> sc.Variable:
...
+            if self._is_time and len(starts) == 0:
+                starts.append(sc.epoch(unit='ns'))
```
jl-wynen, line 180: Is this always correct or only for logs?

src/scippnexus/nxobject.py
```diff
@@ -137,6 +171,18 @@ def __getitem__(self, select) -> sc.Variable:
             self._dataset.read_direct(variable.values, source_sel=index)
         else:
             variable.values = self._dataset[index]
+        if self._is_time or _is_time(variable):
+            starts = []
+            for name in self.attrs:
+                if (dt := _as_datetime(self.attrs[name])) is not None:
+                    starts.append(dt)
+            if self._is_time and len(starts) == 0:
```
jl-wynen, line 179: Why check `self._is_time` again?

src/scippnexus/nxobject.py
```diff
@@ -137,6 +171,18 @@ def __getitem__(self, select) -> sc.Variable:
...
+                starts.append(sc.epoch(unit='ns'))
+            if len(starts) == 1:
```
jl-wynen, line 181: What about >1?

src/scippnexus/docs/our-interpretation-of-the-nexus-format.md
SimonHeybrock: That would be double negation?

src/scippnexus/nxobject.py
SimonHeybrock: I used that initially, but flake8 did not like it.

src/scippnexus/docs/our-interpretation-of-the-nexus-format.md
jl-wynen: Oh, I somehow misread the 'no'.

src/scippnexus/nxobject.py
SimonHeybrock: Oh, wait, maybe it would work with the `r` prefix?

src/scippnexus/nxobject.py
SimonHeybrock: I think so. Do you have a better suggestion?

src/scippnexus/nxobject.py
SimonHeybrock: Because child classes may declare that a field is a datetime field. This handles the case of an absent datetime attribute, and then defaults to using epoch.

src/scippnexus/nxobject.py
jl-wynen: Is there a list of possible attr names?

src/scippnexus/nxobject.py
SimonHeybrock: This is only used when `_is_time` is set manually, which is currently only done by `NXlog`. That is, it would indeed *not* be always correct, and is thus not used by default.

src/scippnexus/nxobject.py
SimonHeybrock: We abort, i.e., only time differences are loaded. See also corresponding test and docs.

src/scippnexus/nxobject.py
SimonHeybrock: No. `start` is used by NXlog/time and `offset` by NXevent_data/event_time_zero. Beyond that nothing is specified.

src/scippnexus/nxobject.py
jl-wynen: Should we warn about it?
Or maybe just log at INFO level. Otherwise, it will be difficult to debug for users.

SimonHeybrock: Based on the discussion earlier, we should *not* strip the info, but convert to UTC (and document that we do that)?

jl-wynen: I think that is the safest approach. Especially if there are datetimes with different timezones in the same file.

SimonHeybrock:
This should be ready for another look. Both the parsing and tests have changed significantly, so check carefully.

tests/nexus_test.py
```diff
@@ -212,6 +212,63 @@ def test_field_of_extended_ascii_in_ascii_encoded_dataset_is_loaded_correctly():
...
+        nxroot['mytime'][...],
+        sc.datetimes(
+            dims=['dim_0'],
+            unit='ns',
+            values=['1970-01-01T00:00:00.000000000', '1970-01-01T00:00:00.000000001']))
+
+
+def test_second_field_with_ns_datetime_attribute_loaded_as_ns_datetime(nxroot):
+    nxroot['mytime'] = sc.arange('ignored', 2, unit='s')
+    nxroot['mytime'].attrs['start_time'] = '1970-01-01T00:00:00.000000000'
```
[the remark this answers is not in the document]
SimonHeybrock: Changed in all but one test, where rounding errors make assertions harder otherwise.

SimonHeybrock:
Had to rewrite the timezone logic again, since the previous edition caused rounding errors.

Aside from that, I configured pytest to treat warnings as errors.

src/scippnexus/nxobject.py
```diff
@@ -84,21 +87,61 @@ def __getitem__(self, name: str) -> Any:
...
+
+def _as_datetime(obj: Any):
+    if isinstance(obj, str):
+        try:
+            # NumPy and scipp cannot handle timezone information. We therefore apply it,
+            # i.e., convert to UTC.
+            # Would like to use dateutil directly, but with Python's datetime we do not
+            # get nanosecond precision. Therefore we combine numpy and dateutil parsing.
+            date_only = 'T' in obj
+            if date_only:
```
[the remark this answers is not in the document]
SimonHeybrock: Had the condition reverted, fixed now.

src/scippnexus/nxobject.py
```diff
@@ -84,21 +88,67 @@ def __getitem__(self, name: str) -> Any:
...
+                # No timezone, parse directly (scipp based on numpy)
+                return sc.datetime(f'{date}T{time}')
+            else:
+                # There is timezone info. Parse with dateutil.
+                dt = dateutil.parser.isoparse(obj)
+                dt = dt.replace(microsecond=0)  # handled by numpy
+                dt = dt.astimezone(datetime.timezone.utc)
+                dt = dt.replace(tzinfo=None).isoformat()
+                # We operate with string operations here and thus end up parsing date
+                # and time twice. The reason is the the timezone-offset arithmetic must
```
jl-wynen, line 133:
```suggestion
                # and time twice. The reason is that the timezone-offset arithmetic
```

tests/nexus_test.py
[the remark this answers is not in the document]
jl-wynen: Are there still rounding errors with you  latest change?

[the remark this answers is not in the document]
SimonHeybrock: Yes. I opened scipp/scipp#2531 for this.

SimonHeybrock:
Leaving the test as-is for now. Could change after the next scipp release.

PR #14, opened 2022-04-08
Add nbstripout pre-commit hook

SimonHeybrock:
- Copy config from scipp repo
- Add `nbstripout`. This is slightly different from `nb-clean` used in CI, but has the advantage of being usable as a pre-commit hook. If it works, my plan is to also use this in CI.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,16 @@
 repos:
-  -   repo: https://github.com/pycqa/flake8
-      rev: 3.8.3
-      hooks:
-        - id: flake8
-  -   repo: https://github.com/pre-commit/mirrors-yapf
-      rev: v0.30.0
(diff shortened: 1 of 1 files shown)


jl-wynen:
I assume, you tested the hook?

SimonHeybrock:
Yes. The only thing it did not strip was and empty list of tags. I don't know if nb-clean is doing that though?

jl-wynen:
Does it matter? Can those be introduced accidentally?

What about empty cells?

SimonHeybrock:
Empty cells are removed by the `'--strip-empty-cells'` option.

Empty list of tags appears when you collapse an markdown section. The collapse information gets stripped, but empty list of tags remains.

jl-wynen (approved):
Ok, let's try it

PR #16, opened 2022-04-11
Fix NXdetector for 0-D detector_number field.

SimonHeybrock:
This also matters when shape=(1,) since this is considered equivalent to shape=() in Nexus

### DIFF
--- a/tests/nxdetector_test.py
+++ b/tests/nxdetector_test.py
@@ -106,8 +106,7 @@ def test_loads_event_data_with_0d_detector_numbers(nxroot):
     assert detector.dims == []
     assert detector.shape == []
     loaded = detector[...]
-    assert sc.identical(loaded.bins.size().data,
-                        sc.array(dims=[], unit=None, dtype='int64', values=2))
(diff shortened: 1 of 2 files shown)


tests/nxdetector_test.py
```diff
@@ -99,6 +99,17 @@ def test_loads_event_data_mapped_to_detector_numbers_based_on_their_event_id(nxr
...
 
+def test_loads_event_data_with_0d_detector_numbers(nxroot):
+    detector = nxroot.create_class('detector0', NX_class.NXdetector)
+    detector.create_field('detector_number', sc.index(1, dtype='int64'))
+    create_event_data_ids_1234(detector.create_class('events', NX_class.NXevent_data))
+    assert detector.dims == []
+    assert detector.shape == []
+    loaded = detector[...]
+    assert sc.identical(loaded.bins.size().data,
+                        sc.array(dims=[], unit=None, dtype='int64', values=2))
```
jl-wynen, line 110: This is a bit odd. Why didn't you use `sc.scalar`?

PR #19, opened 2022-04-11
Some typing improvements

SimonHeybrock:
There are a lot of remaining mypy errors. Not sure what to do with most of them, for now.

### DIFF
--- a/src/scippnexus/nxdata.py
+++ b/src/scippnexus/nxdata.py
@@ -159,7 +159,7 @@ class NXdata(NXobject):
                 continue
             try:
                 sel = to_child_select(self.dims, field.dims, select)
-                coord = self[name][sel]
+                coord: sc.Variable = self[name][sel]
                 for suffix in self._error_suffixes:
                     if f'{name}{suffix}' in self:
                         if coord.variances is not None:
--- a/src/scippnexus/typing.py
+++ b/src/scippnexus/typing.py
@@ -62,8 +62,6 @@ if TYPE_CHECKING:
 
     class ellipsis(Enum):
         Ellipsis = "..."
-
-    Ellipsis = ellipsis.Ellipsis
 else:
     ellipsis = type(Ellipsis)
 
--- a/src/scippnexus/typing.py
+++ b/src/scippnexus/typing.py
@@ -3,6 +3,7 @@
 # @author Simon Heybrock
 from __future__ import annotations
 from typing import Any, Protocol, Union, Tuple, Dict, List, Callable
+from typing import TYPE_CHECKING
 
 
 class H5Base(Protocol):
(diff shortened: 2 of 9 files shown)


src/scippnexus/nxdata.py
```diff
@@ -159,7 +159,7 @@ def _getitem(self, select: ScippIndex) -> sc.DataArray:
                 continue
             try:
                 sel = to_child_select(self.dims, field.dims, select)
-                coord = self[name][sel]
+                coord: sc.Variable = self[name][sel]
```
nvaytet, line 162: Not sure I see why this is required here (and other instances below). In what cases do we need to do this?
Do we alsso need it everywhere in scipp/neutron/ess?

SimonHeybrock: The problem (and the reason for many of the *remaining* errors that I did not manage (or care) to fix yet) is that `__getitem__` may returns different things. Simplest example: `h5py.Group.__getitem__` may give you a `h5py.Group` or a `h5py.Dataset`.

Do you know a solution?

SimonHeybrock: On the contrary, that is the *source of the problem*. If I no do anything with the return value I get complaints that the method may not exist, or follow-up errors about the type of the next `__getitem__`.

nvaytet: Does this mean that every time we get an item from a python dict somewhere, we need to specify the type of what we are expecting?

jl-wynen: > Does this mean that every time we get an item from a python dict somewhere, we need to specify the type of what we are expecting?

Depends on how the dict is defined. If it is e.g. `Dict[str, Union[int, str, tuple]]`, then yes, we would have to check every usage. And strictly speaking that makes sense because we don't know what we are getting. Usually we rely on Python's duck typing and getting `TypeError` or similar exceptions when the type does not match our expectations.

But typically, we would have a more concrete type in the dict: `Dict[str, str]` or any: `Dict[str, Any]` or just `dict`. In those cases, type checking is disabled.

src/scippnexus/typing.py
```diff
@@ -1,24 +1,73 @@
...
+        """Apply callable to all items, recursively"""
+
+
+if TYPE_CHECKING:
+    from enum import Enum
+
+    class ellipsis(Enum):
+        Ellipsis = "..."
 
+    Ellipsis = ellipsis.Ellipsis
```
jl-wynen, line 66: Did you  mean lowercase `ellipsis`? Right now, it conflicts the else branch.

SimonHeybrock: In what way? We need `ellipsis` to be a type (the class defined here). I got this from https://github.com/python/typing/issues/684#issuecomment-548203158.

jl-wynen: Looks to me like `Ellipsis` is only defined in order to support `if a is Ellipsis:` in the example. In our case, writing 'Ellipsis' in code outside of `typing.py` would refer to the builtin type instead. So I am not sure the type checker would be happy with it.
In any case, I think line 66 can be removed in our case.

jl-wynen (approved):

PR #20, opened 2022-04-13
Fix and improve unit handling in time-dependent transformation chains

SimonHeybrock:
- Add tests for existing working behavior (chains with different value units).
- Add test for chains with time dependence (same unit, working before).
- Add test for chains with time dependence and different unit. This was broken.
- Fix for previous item, and avoid forcing things to `nanoseconds`. Instead use unit of offsets, and then continue with smallest unit if there are multiple transforms in the chain.

This is required to fix some tests in `scippneutron`.

### DIFF
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -187,13 +187,13 @@ class Field:
             self._dataset.read_direct(variable.values, source_sel=index)
         else:
             variable.values = self._dataset[index]
-        if self._is_time or _is_time(variable):
+        if _is_time(variable):
             starts = []
             for name in self.attrs:
                 if (dt := _as_datetime(self.attrs[name])) is not None:
                     starts.append(dt)
             if self._is_time and len(starts) == 0:
-                starts.append(sc.epoch(unit='ns'))
+                starts.append(sc.epoch(unit=self.unit))
             if len(starts) == 1:
                 variable = convert_time_to_datetime64(
                     variable,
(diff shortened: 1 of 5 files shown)


SimonHeybrock:
Note that we now depend on scipy, since we use interpolation from scipp. I am not all too happy about this. Is there a better solution?

src/scippnexus/nxobject.py
```diff
@@ -187,13 +187,13 @@ def __getitem__(self, select) -> sc.Variable:
             self._dataset.read_direct(variable.values, source_sel=index)
         else:
             variable.values = self._dataset[index]
-        if self._is_time or _is_time(variable):
```
jl-wynen, line 190: Why is `self._is_time` no longer required / was wrong?

SimonHeybrock: Previously this would force handling as time, even if the unit was not a time unit. You would thus get an UnitError when trying to add a base datetime. I think it just caused unnecessary trouble doing so.

src/scippnexus/nxtransformations.py
```diff
@@ -90,6 +90,18 @@ def _interpolate_transform(transform, xnew):
                                    fill_value="extrapolate")(xnew=xnew)
 
 
+def _smaller_unit(a, b):
+    if a.unit == b.unit:
+        return a.unit
+    a = sc.scalar(1.0, unit=a.unit)
+    b = sc.scalar(1.0, unit=b.unit)
+    ratio = a / b
+    if ratio.value < 1.0:
```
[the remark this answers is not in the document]
SimonHeybrock: You comment also highlights that my code was not doing what I wanted it to do... changing.

jl-wynen (approved):

PR #25, opened 2022-04-26
Freeze requirements that are pip-installed during CI and release builds

SimonHeybrock:
Let us freeze more!


.github/workflows/ci.yml
```diff
@@ -16,8 +16,7 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: 3.8
-      - run: python -m pip install --upgrade pip
```
jl-wynen, line 19: Are we guaranteed to already have a suitable version of pip?

SimonHeybrock: I don't really know, but what I do now is as shown in an example of the `actions/setup-python`, so I'd hope so?

SimonHeybrock: Other docs (https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python) *do* still have the `--upgrade pip` call, so I am not sure what to make of this...

SimonHeybrock: I added it back in.

jl-wynen (approved):
Ok. Maybe it does not make a significant difference. The action might be only be a few weeks or months behind.

PR #32, opened 2022-06-01
Only squeeze dimensions that are beyond the list of dims specified in Field

nvaytet:


src/scippnexus/nxobject.py
```diff
@@ -161,6 +161,8 @@ def __init__(self, dataset: H5Dataset, dims=None, is_time=None):
             self._dims = dims
             if len(self._dims) < len(self._shape):
                 self._shape = [size for size in self._shape if size != 1]
+                # If axes are partially specified, we may have squeezed too many dims
+                self._shape = ([1] * (len(self._dims) - len(self._shape))) + self._shape
```
jl-wynen, line 165: How do you know that the order is correct?

SimonHeybrock (changes requested):

src/scippnexus/nxobject.py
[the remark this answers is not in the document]
nvaytet: Yes, I am not sure I grasped all the logic.

So how do you know that you need to put the `time` as the first element here? https://github.com/scipp/scippnexus/blob/main/src/scippnexus/nxlog.py#L34

Because of the comment "The outermost axis in NXlog is pre-defined to 'time' (if present)." ?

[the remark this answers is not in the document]
nvaytet: In your example above, if you have `shape=[1,2,1,1]` and `dims=['x', 'y']`, how do you know what's what?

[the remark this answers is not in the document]
SimonHeybrock: I think we either need to raise in all cases, or adopt a convention (e.g., dims apply from the left). But swapping things around is broken regardless of the conversion, isn't it?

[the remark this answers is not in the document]
jl-wynen: Does the file contain enough information to name all dimensions? In that case, we could pass the full list of dims and shape to `Field` and squeeze length-1 dims only after processing.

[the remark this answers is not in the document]
SimonHeybrock: > Does the file contain enough information to name all dimensions?

No. It is Nexus, so the answer is no, you have to guess.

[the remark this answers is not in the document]
nvaytet: Maybe we should just not squeeze?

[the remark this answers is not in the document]
SimonHeybrock: That is not an option, because Nexus treats `shape=()` and `shape=(1,)` interchangeably.

src/scippnexus/nxdata.py
```diff
@@ -116,24 +116,27 @@ def _guess_dims(self, name: str):
...
+            return self._try_guess_dims(name)
         if name in self._get_axes():
             # If there are named axes then items of same name are "dimension
             # coordinates", i.e., have a dim matching their name.
             return [name]
-        try:
-            return self._guess_dims(name)
-        except NexusStructureError:
-            return None
+        return self._try_guess_dims(name)
```
SimonHeybrock, line 139: Why moved to function?

nvaytet: Because I am using it in two places: here and 5 lines above.

src/scippnexus/nxobject.py
```diff
@@ -160,11 +160,18 @@ def __init__(self, dataset: H5Dataset, dims=None, is_time=None):
         if dims is not None:
             self._dims = dims
             if len(self._dims) < len(self._shape):
-                self._shape = [size for size in self._shape if size != 1]
+                self._shape = [size for size in self._shape if size > 1]
```
SimonHeybrock, line 163: Why is this dropping 0s now?

src/scippnexus/nxobject.py
```diff
@@ -160,11 +160,18 @@ def __init__(self, dataset: H5Dataset, dims=None, is_time=None):
...
+                    f"Dims: {self._dims} and shape: {self._shape} do not have "
+                    "the same length.")
         elif (axes := self.attrs.get('axes')) is not None:
             self._dims = axes.split(',')
+            if len(self._dims) != len(self._shape):
+                raise ValueError(
+                    "The number of axes is different from the number of sizes.")
         else:
-            self._shape = [size for size in self._shape if size != 1]
+            self._shape = [size for size in self._shape if size > 1]
```
SimonHeybrock, line 174: Why dropping 0s?

tests/nxdata_test.py
```diff
@@ -316,6 +316,53 @@ def test_unnamed_extra_dims_of_multidim_coords_are_squeezed(nxroot):
...
+
+def test_raises_when_only_one_axis_specified(nxroot):
+    signal = sc.array(dims=['xx', 'yy'], unit='m', values=[[1.1]])
+    data = nxroot.create_class('data1', NX_class.NXdata)
+    data.create_field('signal', signal)
+    data.attrs['axes'] = ['xx']
+    data.attrs['signal'] = 'signal'
+    with pytest.raises(ValueError):
+        _ = data[...]
+
```
SimonHeybrock, line 364: Won't we run into this systematically with `NXlog`, where we hard-code the outer axis?

src/scippnexus/nxobject.py
```diff
@@ -201,7 +201,13 @@ def __getitem__(self, select) -> sc.Variable:
...
             variable.values = np.asarray(strings).flatten()
         elif variable.values.flags["C_CONTIGUOUS"]:
-            self._dataset.read_direct(variable.values, source_sel=index)
+            try:
+                self._dataset.read_direct(variable.values, source_sel=index)
+            except TypeError:
+                # A TypeError occurs in some cases where h5py cannot broadcast data
+                # with e.g. shape (20, 1) to a buffer of shape (20,).
+                # Note that broadcasting (1, 20) -> (20,) does work.
+                variable.values = self._dataset[index].reshape(shape)
```
SimonHeybrock, line 207: This looks very dangerous. It could hide some potentially really bad bugs.

nvaytet: So I tried for a while yesterday to add an extra slice to `source_sel` if there were some dimensions that needed squeezing, but it was simply not working.

As it turns out, I just tried the simple example again at home last night, and it magically worked:
```Py
import numpy as np
import h5py

with h5py.File('my_hdf5_file.h5', 'w') as f:
    dset = f.create_dataset("dset", (5, 2), dtype='int64')
    arr = np.ones((5,), dtype='int64')
    dset.read_direct(arr, np.s_[:, 0])
    print(arr)
```

Difference: `h5py=3.1.0` at work (because of the `<3.2` pinning due to mantid), `h5py=3.4.0` at home :-(

SimonHeybrock: Can you find out if that is a deliberate fix on their side? If so, then we could bump the required version.

nvaytet: This? https://github.com/h5py/h5py/pull/1796

SimonHeybrock: 👍 can you add a comment with the h5py version in our code?

src/scippnexus/nxobject.py
```diff
@@ -180,15 +185,31 @@ def __getitem__(self, select) -> sc.Variable:
                             shape=shape,
                             dtype=self.dtype,
                             unit=self.unit)
+
+        # If the variable is empty, return early
+        if np.prod(shape) == 0:
+            return variable
+
```
SimonHeybrock, line 192: Why is this necessary?

src/scippnexus/nxobject.py
```diff
@@ -180,15 +185,31 @@ def __getitem__(self, select) -> sc.Variable:
...
+            # of shape (20,). Note that broadcasting (1, 20) -> (20,) does work.
+            # Therefore, we manually squeeze here.
+            # A pin of h5py<3.2 is currently required by Mantid and hence scippneutron
+            # (see https://github.com/h5py/h5py/issues/1880#issuecomment-823223154)
+            # hence this workaround. Once we can use a more recent h5py with Mantid,
+            # this try/except can be removed.
+            try:
+                self._dataset.read_direct(variable.values, source_sel=index)
+            except TypeError:
+                variable.values = self._dataset[index].squeeze()
```
SimonHeybrock, line 212: Does `h5py.Dataset.squeeze` trigger a load if the data? If not, i.e., if it is cheap, then I don't the the point of `try`/`except` --- just always squeeze, and avoid most of the lengthy comment (and more importantly the "control-flow-via-exceptions" anti-pattern)?

nvaytet: I don't think `h5py.Dataset.squeeze` exists?
The point of the try/except is that I think when you use `read_direct` you are not making a copy, while when assigning the `self._dataset[index]` to the `.values` you are making a copy? So better to try and use `read_direct` if you can.
But maybe I understood wrong?

src/scippnexus/nxobject.py
nvaytet: I was getting errors in `read_direct` when trying to read in a NXlog with shape 0.
And I thought since there is nothing to load anyway, I could just return the empty variable early?

src/scippnexus/nxobject.py
SimonHeybrock: No that is fine. If it does not exist then what you do is the only option.

src/scippnexus/nxobject.py
SimonHeybrock: 👍 I vaguely remember I had this is with `read_direct` before.

PR #33, opened 2022-06-01
Fixes for scipp-0.14

SimonHeybrock:
Will make a patch release when merged.


tests/nxdetector_test.py
```diff
@@ -208,8 +208,8 @@ def test_loading_event_data_with_full_selection_and_automatic_detector_numbers_w
     detector = nxroot.create_class('detector0', NX_class.NXdetector)
     create_event_data_ids_1234(detector.create_class('events', NX_class.NXevent_data))
     assert detector.dims == ['detector_number']
-    assert detector[...].shape == [4]
-    assert detector[()].shape == [4]
+    assert tuple(detector[...].shape) == (4, )
+    assert tuple(detector[()].shape) == (4, )
```
jl-wynen, line 212: Looks odd. Why does this convert the LHS to a tuple and then test with a tuple on the RHS? Looks like the LHS should have already been a tuple to warrant a change.

jl-wynen (approved):

PR #37, opened 2022-06-21
Avoid deprecation warnings in scipp dev builds and next release

SimonHeybrock:
Warnings from deprecated syntax of `sc.bin` got very noise when loading files (with many banks). This should avoid this, and also still works with older versions.

### DIFF
--- a/src/scippnexus/nxdetector.py
+++ b/src/scippnexus/nxdetector.py
@@ -10,6 +10,13 @@ from .nxdata import NXdata
 from .nxevent_data import NXevent_data
 
 
+def group(da: sc.DataArray, groups: sc.Variable) -> sc.DataArray:
+    if hasattr(da, 'group'):
+        return da.group(groups)
+    else:
+        return sc.bin(da, groups=[groups])
+
+
 class EventSelector:
     """A proxy object for creating an NXdetector based on a selection of events.
     """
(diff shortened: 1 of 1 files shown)


src/scippnexus/nxdetector.py
```diff
@@ -10,6 +10,13 @@
 from .nxevent_data import NXevent_data
 
 
+def group(da: sc.DataArray, groups: sc.Variable) -> sc.DataArray:
+    if hasattr(da, 'group'):
+        return da.group(groups)
+    else:
+        return sc.bin(da, groups=[groups])
```
jl-wynen, line 17: How do we make sure to remove this wrapper once we no longer need it?

SimonHeybrock: There is no real harm to it, I suppose? It depends on how long we want to support older versions of scipp?

jl-wynen: To me it was mainly a question about clean, concise code.

> It depends on how long we want to support older versions of scipp?

Don't know. But we still have a small userbase and only a pre-release. So I would expect to drop support for this version eventually.

SimonHeybrock: 👍 
Having had this discussion will serve as a reminder ;)

jl-wynen (approved):

PR #39, opened 2022-07-04
Update action syntax in docs.yml

SimonHeybrock:
See https://github.blog/changelog/2022-06-10-github-actions-inputs-unified-across-manual-and-reusable-workflows/.

Do we still need the `${{ }}`?

Edit: Now also including changes for docs publish folders.


SimonHeybrock:
Reverted changes to actual workflows, added a new one for manual testing (need to merge).

.github/workflows/docs.yml
```diff
@@ -30,23 +35,41 @@ jobs:
           python-version: 3.8
       - run: python -m pip install --upgrade pip
       - run: python -m pip install -r requirements/ci.txt
-      - run: tox --skip-pkg-install -e docs scippnexus==${GITHUB_REF_NAME}
-        if: ${{ inputs.publish && inputs.release }}
-      - run: tox --skip-pkg-install -e docs scippnexus
-        if: ${{ inputs.publish && !inputs.release }}
+      - run: tox --skip-pkg-install -e docs scippnexus==${VERSION}
```
jl-wynen: By the way, we are not running linkcheck. Is this intentional?

PR #41, opened 2022-07-15
Docs build fixes

### LINKED ISSUE #31: Docs workflow
See #30 and recent actions.

This is almost good to let us publish or update docs. One remaining subtlety:

- If we make docs of an old branch, e.g., old tag, we may want to use an old package. So we should extend this to not simply take a `publish` flag, but a version number? Then this can be used to:
  1. Get correct PyPI package (or use local one if `dev` is specified).
  2. Publish to correct folder (including version tag).
  
  This would also need to be able to handle extra folders, and specifying 'latest', to just use latest package (and publish into root).

SimonHeybrock:
Addressing comments from #40.

Fixes #31.


jl-wynen:
I'm happy with it.

But can you add a short README to explain how the release works and where docs are stored? It might be hard to understand from the workflows in the future.

SimonHeybrock:
> I'm happy with it.
> 
> But can you add a short README to explain how the release works and where docs are stored? It might be hard to understand from the workflows in the future.

I suggest I do so in `scipp` when we know this really works.

jl-wynen:
> I suggest I do so in scipp when we know this really works.

There should still be at least a file with a link in all packages. Otherwise, one needs to know where to look.

PR #45, opened 2022-08-04
Dynamic properties based on classes found in group

SimonHeybrock:
This is the most basic (and uncontroversial) part of what we discussed as part of scipp/scippneutron#359. This dynamically "adds" properties (by defining `__getattr__`), based on which NX_class instances are found in the current group (scippnexus.NXobject).

The code is *not* meant for review yet (a couple things to fix), for now please just try this out and comment on:

- Is this functionality what we want?
- Does the tab-completion work and make sense to you?
- etc.

To test:

```python
import scippnexus as snx
from scippnexus import data
path = data.get_path('PG3_4844_event.nxs')
f = snx.File(path)
```
then try things such as:

```
f.entry
f.entry.sample
f.entry.instrument.detector
f.entry.instrument.detectors
```

### DIFF
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -29,7 +29,6 @@ class DimensionedArray(Protocol):
 
     Could be, e.g., a scipp.Variable or a dimple dataclass wrapping a numpy array.
     """
-
     @property
     def values(self):
         """Multi-dimensional array of values"""
@@ -44,7 +43,6 @@ class DimensionedArray(Protocol):
 
 
 class AttributeManager(Protocol):
-
     def __getitem__(self, name: str):
         """Get attribute"""
 
@@ -73,7 +71,6 @@ class NX_class(Enum):
 class Attrs:
     """HDF5 attributes.
     """
-
     def __init__(self, attrs: AttributeManager):
         self._attrs = attrs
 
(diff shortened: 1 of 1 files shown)


SimonHeybrock:
Is it a bad idea to use the "short" class names? There may be potential clashes with "normal" properties (see https://docs.h5py.org/en/stable/high/group.html#reference or potentially https://docs.h5py.org/en/stable/high/dataset.html#reference for what we may end up with eventually. Should we instead prefix with "nx", i.e,

```python
f.nxentry.nxinstrument.nxdetectors
f.nxentry.nxmonitors
```
?

This would however make it less likely that we would have a symmetric interface for an in-memory data-reduction-specific class --- the latter would not care about Nexus, users simply want "the monitors".

jl-wynen:
Some tests of autocomplete:
In Jupyter with packages
```.yaml
  - jupyterlab-lsp
  - python-lsp-server
  - pip:
      - pyls-mypy
```
I don't get any automatic autocomplete suggestions as earlier. I have to press tab + enter to get something completed (including static attributes). But it actually crashed the language server and no completions worked afterwards.

And mypy does not report any diagnostics. But it appears to not have any data on scippnexus. Need to investigate this independently.

Pycharm does not show any diagnostics for `File`. It does not warn about using wrong attributes and only has autocompletions for static attibutes.

SimonHeybrock:
Will be replaced by #49.

PR #46, opened 2022-08-23
Update logo and use SVG for docs instead of converting to PNG

SimonHeybrock:


jl-wynen (approved):

nvaytet:
Why are there two logo files and why is one much smaller than the other? Is it because one uses fonts and the other paths?

SimonHeybrock:
> Why are there two logo files and why is one much smaller than the other? Is it because one uses fonts and the other paths?

- Yes that is why there are two files. We keep the font version to be able to change in the future.
- Biggest reason for size difference is that the path version is plain SVG, instead of Inkscape SVG.

PR #48, opened 2022-08-25
Switch to using classes as keys and add child selection based on NX_class

SimonHeybrock:
This is part of an effort to improve the overall interface. #45 experiments with providing properties for accessing specific children based on their class, without requiring knowledge of their name. As reported in https://github.com/scipp/scippneutron/issues/359#issuecomment-1223869619 this has some issues. In particular the singular/plural "overloading" does not appear to be a workable solution. Instead we will:

- Provide attributes for accessing unique child, e.g., `file.entry.instrument` (this will be in a follow-up PR, basically a modification of #45).
- Access "all (direct) children of certain class" by providing `__getitem__` that accepts the NX_class as a key (this PR).

Additional changes here:
- The enum `NX_class` is removed. I have always felt that it was cumbersome to use. The second bullet above means users will need to use such keys more (instead of group names). Therefore we switch to using the classes as keys.
- Remove `by_nx_class`, which had a number of problems (see commit message). This is used in `scippnexus.load_nexus`, which will need refactoring.

### DIFF
--- a/src/scippnexus/nxdetector.py
+++ b/src/scippnexus/nxdetector.py
@@ -20,6 +20,7 @@ def group(da: sc.DataArray, groups: sc.Variable) -> sc.DataArray:
 class EventSelector:
     """A proxy object for creating an NXdetector based on a selection of events.
     """
+
     def __init__(self, detector):
         self._detector = detector
 
@@ -36,6 +37,7 @@ class _EventField:
     This has no equivalent in the NeXus format, but represents the conceptual
     event-data "signal" dataset of an NXdetector.
     """
+
     def __init__(self,
                  nxevent_data: NXevent_data,
                  event_select: ScippIndex,
@@ -115,6 +117,7 @@ class NXdetector(NXobject):
     is used to map event do detector pixels. Otherwise this returns event data in the
     same format as NXevent_data.
     """
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._event_select = tuple()
--- a/src/scippnexus/nxdetector.py
+++ b/src/scippnexus/nxdetector.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from copy import copy
 from typing import List, Optional, Union
 import scipp as sc
-from .nxobject import NX_class, NXobject, Field, ScippIndex, NexusStructureError
+from .nxobject import NXobject, Field, ScippIndex, NexusStructureError
 from .nxdata import NXdata
 from .nxevent_data import NXevent_data
 
@@ -20,7 +20,6 @@ def group(da: sc.DataArray, groups: sc.Variable) -> sc.DataArray:
 class EventSelector:
     """A proxy object for creating an NXdetector based on a selection of events.
     """
-
     def __init__(self, detector):
         self._detector = detector
 
@@ -37,7 +36,6 @@ class _EventField:
     This has no equivalent in the NeXus format, but represents the conceptual
     event-data "signal" dataset of an NXdetector.
     """
-
     def __init__(self,
                  nxevent_data: NXevent_data,
                  event_select: ScippIndex,
@@ -117,7 +115,6 @@ class NXdetector(NXobject):
     is used to map event do detector pixels. Otherwise this returns event data in the
     same format as NXevent_data.
     """
-
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._event_select = tuple()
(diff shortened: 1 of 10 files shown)


src/scippnexus/nxdetector.py
```diff
@@ -170,7 +170,7 @@ def events(self) -> Union[None, NXevent_data]:
         # The standard is unclear on whether the 'data' field may be NXevent_data or
         # whether the fields of NXevent_data should be stored directly within this
         # NXdetector. Both cases are observed in the wild.
-        event_entries = self.by_nx_class()[NX_class.NXevent_data]
+        event_entries = self[NXevent_data]
```
jl-wynen, line 173: Are there files where this changes the result?

src/scippnexus/nxobject.py
```diff
@@ -320,9 +305,18 @@ def _get_child(
...
+    def _get_children_by_nx_class(self, select: type) -> Dict['NXobject']:
+        children = {}
+        for key in self.keys():
+            if (child := self._get_child(key)).attrs.get('NX_class') == select.__name__:
+                children[key] = child
+        return children
+
     def __getitem__(
             self,
             name: NXobjectIndex) -> Union['NXobject', Field, sc.DataArray, sc.Dataset]:
```
jl-wynen, line 317:
```suggestion
            name: Union[NXobjectIndex, type]) -> Union['NXobject', Field, sc.DataArray, sc.Dataset, Dict['Nxobject']]:
```
Or cleaner, use https://docs.python.org/3/library/typing.html#typing.overload

src/scippnexus/nxobject.py
```diff
@@ -320,9 +305,18 @@ def _get_child(
             da.coords['depends_on'] = t if isinstance(t, sc.Variable) else sc.scalar(t)
         return da
 
+    def _get_children_by_nx_class(self, select: type) -> Dict['NXobject']:
+        children = {}
+        for key in self.keys():
```
jl-wynen, line 310: Could this use `self.items()`?

src/scippnexus/nxobject.py
```diff
@@ -320,9 +305,18 @@ def _get_child(
...
+            if (child := self._get_child(key)).attrs.get('NX_class') == select.__name__:
```
jl-wynen, line 311: Compare using `child.nx_class`?

src/scippnexus/nxdetector.py
SimonHeybrock: I doubt it. It would mean you'd have a structure like `NXdetector/NXsomething/NXevent_data` instead of `NXdetector/NXevent_data`. I have never seen that, and the Nexus standard does not suggest it should exist.

src/scippnexus/nxobject.py
SimonHeybrock: No, because that will call `self._get_child` with `use_field_dims=True`, leading to infinite recursion.

src/scippnexus/nxobject.py
SimonHeybrock: I was wondering about that: Should `nx_class` return the instance, or what is in the attribute? There are some special cases where these differ. Do we actually want/need this attribute at all?

SimonHeybrock: Not return an instance, but the instance may differ from what the attribute defines. The special cases are things such as NXmonitor and NXlog, which are conceptually treated like NXdata during load (and implemented by deferring to that). This is not exposed to the user, so not really relevant.

SimonHeybrock: I'll keep it, as it may be useful for implementing things such as "give me all children with the same nxclass as this group".

src/scippnexus/nxobject.py
```diff
@@ -305,16 +305,48 @@ def _get_child(
...
+    def __getitem__(self, name):
+        """
+        Get a child group or child dataset, a selection of child groups, or load and
+        return the current group.
+
+        Three cases are supported:
+
+        - String name: The child group or child dataset of that name is returned.
+        - Class such as ``NXdata`` or ``NXlog``: A dict containing all direct children
+          with a matching ``NX_class`` attribute are returned.
```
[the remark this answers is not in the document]
SimonHeybrock: I don't think this was *the* initial idea. `obj[NXdata, 'data1']` was an option we considered briefly, but it does not work well because it required knowledge of the group name, which is *very* inconvenient in practice. You can now to `obj[NXdata]['data1']` is you want to achieve the same (and in principle we could add an overload for this (but maybe wait until we get some actual usage experience?)).

jl-wynen (approved):

PR #49, opened 2022-08-25
Dynamic properties based on unique classes found in group

SimonHeybrock:
This supersedes #45. Here we just add properties for the "unique" child groups, avoiding conceptual complications and complicated code. Combined with #48 this should provide a more convenient interface.

To do after positive initial review:
- [x] Unit tests
- [x] Release notes

### DIFF
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -454,7 +454,6 @@ class NXobject:
         raise RuntimeError(f"Multiple keys match {nxclass}, use obj[{nxclass}] to "
                            f"obtain all matches instead of obj.{attr}.")
 
-    @functools.lru_cache(maxsize=1)
     def __dir__(self):
         nxclasses = []
         for val in self.values():
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -466,13 +466,18 @@ class NXobject:
                                   f"to obtain all matches instead of obj.{attr}.")
 
     def __dir__(self):
-        nxclasses = []
-        for val in self.values():
-            if isinstance(val, NXobject):
-                nxclasses.append(val.nx_class)
         keys = super().__dir__()
+        try:
+            nxclasses = []
+            for val in self.values():
+                if isinstance(val, NXobject):
+                    nxclasses.append(val.nx_class)
+        except Exception:
+            pass
         for key in set(nxclasses):
             if key is None:
+                continue
+            if key in keys:
                 continue
             if nxclasses.count(key) == 1:
                 keys.append(key.__name__[2:])
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -451,8 +451,8 @@ class NXobject:
             raise NexusStructureError(f"No group with requested NX_class='{nxclass}'")
         if len(matches) == 1:
             return next(iter(matches.values()))
-        raise RuntimeError(f"Multiple keys match {nxclass}, use obj[{nxclass}] to "
-                           f"obtain all matches instead of obj.{attr}.")
+        raise NexusStructureError(f"Multiple keys match {nxclass}, use obj[{nxclass}] "
+                                  f"to obtain all matches instead of obj.{attr}.")
 
     def __dir__(self):
         nxclasses = []
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -484,6 +484,12 @@ class NXobject:
         return keys
 
 
+def _make(group) -> NXobject:
+    if (nx_class := Attrs(group.attrs).get('NX_class')) is not None:
+        return _nx_class_registry().get(nx_class, NXobject)(group)
+    return group  # Return underlying (h5py) group
+
+
 class NXroot(NXobject):
     """Root of a NeXus file."""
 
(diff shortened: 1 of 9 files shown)


jl-wynen (commented):
What happens when the `NXobject` is used after the file was closed? Normally, some functions would fail. But with the dynamic attrs, can it happen that even looking up an attr or using `dir` requires an open file?

> To do after positive initial review:
>
> Unit tests

Please add tests before I approve.

SimonHeybrock:
> What happens when the `NXobject` is used after the file was closed? Normally, some functions would fail. But with the dynamic attrs, can it happen that even looking up an attr or using `dir` requires an open file?

Same as before: You get some error from h5py. Yes looking things up requires an open file, but as far as I can tell NXobject was (and still is) entirely unusable unless the file is open?

src/scippnexus/nxobject.py
```diff
@@ -441,6 +442,32 @@ def __setitem__(self, name: str, value: Union[Field, NXobject, DimensionedArray]
...
+            raise AttributeError(f"'NXobject' object has no attribute {attr}")
+        matches = self[nxclass]
+        if len(matches) == 0:
+            raise NexusStructureError(f"No group with requested NX_class='{nxclass}'")
+        if len(matches) == 1:
+            return next(iter(matches.values()))
+        raise RuntimeError(f"Multiple keys match {nxclass}, use obj[{nxclass}] to "
+                           f"obtain all matches instead of obj.{attr}.")
+
+    @functools.lru_cache(maxsize=1)
```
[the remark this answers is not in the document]
SimonHeybrock: It is just a list of strings so likely not relevant.... but since performance is not crucial here (at least for now) I will remove it.

jl-wynen:
> > What happens when the `NXobject` is used after the file was closed? Normally, some functions would fail. But with the dynamic attrs, can it happen that even looking up an attr or using `dir` requires an open file?
> 
> Same as before: You get some error from h5py. Yes looking things up requires an open file, but as far as I can tell NXobject was (and still is) entirely unusable unless the file is open?

I'm worried that some automatic Python stuff breaks if `__getattr__` and `__dir__` don't work. But I don't know if there is such a case. Are you confident that it is fine?

src/scippnexus/nxobject.py
```diff
@@ -466,14 +466,19 @@ def __getattr__(self, attr: str) -> Union[Any, 'NXobject']:
...
-        for val in self.values():
-            if isinstance(val, NXobject):
-                nxclasses.append(val.nx_class)
         keys = super().__dir__()
+        try:
+            nxclasses = []
+            for val in self.values():
+                if isinstance(val, NXobject):
+                    nxclasses.append(val.nx_class)
+        except Exception:
```
jl-wynen, line 475: Can this be made more concrete? E.g. `(RuntimeError, KeyError)`?

src/scippnexus/nxobject.py
```diff
@@ -519,10 +524,18 @@ def _nx_class_registry():
...
+        for name, cls in inspect.getmembers(nexus_classes, inspect.isclass)
+    }
+    implemented = {
         cls.__name__: cls
         for cls in [
             NXroot, NXentry, NXevent_data, NXlog, NXmonitor, NXdata, NXdetector,
             NXsample, NXsource, NXdisk_chopper, NXinstrument, NXtransformations
         ]
     }
+    assert set(lut).isdisjoint(implemented)
```
jl-wynen, line 539: Is this required? Seems to like it is fine to overwrite with an 'implemented' class. If you want to test that whena  new one is implemented, the empty class is removed, add a test like this to the actual unit tests.

SimonHeybrock: Without this we can end up with defining the same class twice, in different submodules. This can then lead to comparison failed since, e.g., `moduleA.NXdata != moduleB.NXdata`.

src/scippnexus/nxobject.py
SimonHeybrock: I don't know, `values()` instantiates the various specific `NXobject` subclasses. We don't know for sure what can go wrong there?

src/scippnexus/nxobject.py
SimonHeybrock: How? What specifically would you test?

jl-wynen: Copy-paste this function up to the assertion. This would only detect a problem if the list of implemented classes gets updated. But that is the same as with `__dir__`.

src/scippnexus/nxobject.py
jl-wynen: But are those failures reasons to omit the attr or would it be a hard failure that should raise an exception with the user?

src/scippnexus/nxobject.py
SimonHeybrock: I don't think is a solution: As soon as we implement a new class, we should change the test, but we would forget.... and I think we actually have the same problem in this function itself. I think I will try to move the assert into `__init__.py`.

SimonHeybrock: Avoided this mess in an update, as discussed offline.

jl-wynen (approved):

src/scippnexus/nxobject.py
SimonHeybrock: Neither: It most likely indicates a problem about an unrelated child, we should probably try to avoid `values()` here?

SimonHeybrock: @jl-wynen I refactored this, please have a look at the latest commit.

src/scippnexus/nxobject.py
```diff
@@ -441,6 +453,41 @@ def __setitem__(self, name: str, value: Union[Field, NXobject, DimensionedArray]
...
+        raise NexusStructureError(f"Multiple keys match {nxclass}, use obj[{nxclass}] "
+                                  f"to obtain all matches instead of obj.{attr}.")
+
+    def __dir__(self):
+        keys = super().__dir__()
+        nxclasses = []
+        # Avoiding self.values() since it is more costly, but mainly since there may be
+        # edges cases where creation of Field/NXobject may raise on unrelated children.
+        for name, val in self._group.items():
+            if not hasattr(val, 'shape'):  # not a dataset
```
jl-wynen, line 474:
```suggestion
            if isinstance(val, h5.Group):
```
?

src/scippnexus/nxobject.py
```diff
@@ -441,6 +453,41 @@ def __setitem__(self, name: str, value: Union[Field, NXobject, DimensionedArray]
...
+        if len(matches) == 1:
+            return next(iter(matches.values()))
+        raise NexusStructureError(f"Multiple keys match {nxclass}, use obj[{nxclass}] "
+                                  f"to obtain all matches instead of obj.{attr}.")
+
+    def __dir__(self):
+        keys = super().__dir__()
+        nxclasses = []
+        # Avoiding self.values() since it is more costly, but mainly since there may be
+        # edges cases where creation of Field/NXobject may raise on unrelated children.
```
jl-wynen, line 472:
```suggestion
        # edge cases where creation of Field/NXobject may raise on unrelated children.
```

src/scippnexus/nxobject.py
SimonHeybrock: No, because we need this to work with other things such as the JSON interface. Another (not so relevant) example is a wrapper around `tiled`.

jl-wynen: OK

jl-wynen (approved):

PR #53, opened 2022-08-30
Improve NXdetector dim guessing

SimonHeybrock:


src/scippnexus/nxdetector.py
```diff
@@ -158,9 +158,12 @@ def _nxdata(self, use_event_signal=True) -> NXdata:
         # NXdata uses the 'signal' attribute to define the field name of the signal.
         # NXdetector uses a "hard-coded" signal name 'data', without specifying the
         # attribute in the file, so we pass this explicitly to NXdata.
+        # Note the special case of an NXevent_data subgroup named 'data', which we
+        # avoid by checking if 'data' is a dataset.
```
jl-wynen, line 162: Are there files like this? What does the 'data' group contain?

SimonHeybrock: For example the AMOR file. The `data` group contains the event data (i.e., it is NXevent_data).

jl-wynen (approved):

PR #54, opened 2022-09-06
Load all content from NXsource, NXsample, and NXdisk_chopper

SimonHeybrock:
Based on a chat with TR: Apparently NXlog groups are meant for replacing arbitrary fields. That is, instead of loading a list of NXlog groups from somewhere in the file, we should instead simply load possible parent groups in a way that supports this.

This implies in particular that groups such as NXsample may contain, e.g., a `temperature` that is not a 0-D field, but yields a DataArray. We can thus not assume the NXsample can be represented as `scipp.Dataset`. I am therefore switching to `dict` here. Another alternative would have been wrapping in scalar variables, but I do not see an advantage on this level.

### DIFF
--- a/src/scippnexus/leaf.py
+++ b/src/scippnexus/leaf.py
@@ -1,0 +1,25 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+# @author Simon Heybrock
+from typing import Dict, Union
+import scipp as sc
+from .nxobject import NXobject, ScippIndex
+from ._common import to_plain_index
+
+
+class Leaf(NXobject):
+    """Base class for "leaf" groups than can be loaded as a dict.
+    """
+
+    def _getitem(self,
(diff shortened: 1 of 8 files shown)


src/scippnexus/leaf.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Leaf(NXobject):
+    """Base class for "leaf" groups than can be loaded as a dict.
+    """
+
+    def _getitem(self,
+                 select: ScippIndex) -> Dict[str, Union[sc.Variable, sc.DataArray]]:
+        from .nexus_classes import NXtransformations
+        index = to_plain_index([], select)
+        if index != tuple():
```
nvaytet, line 18: Not sure I followed everything but `index` doesn't seem to be used except for this check. Is that intended?

SimonHeybrock: `__getitem__` uses different key types to do different things. `str` to get a child, others for loading a slice. Since these objects to not have a definite shape we only support "loading everything", which you can do by passing an empty tuple.

nvaytet (approved):

PR #56, opened 2022-09-08
Use list instead of tuple of classes in __getitem__

SimonHeybrock:
A list is more consistent with other Python behavior. For example, in NumPy a tuple is used for multi-dim slicing, but a list can select multiple indices along a single dimension: `a[[2,5]]`.

### DIFF
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -320,8 +320,9 @@ class NXobject:
 
     def _get_children_by_nx_class(
             self, select: Union[type,
-                                Tuple[type]]) -> Dict[str, Union['NXobject', Field]]:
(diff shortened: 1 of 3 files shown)


src/scippnexus/nxobject.py
```diff
@@ -368,7 +369,7 @@ def __getitem__(self, name):
         def isclass(x):
             return inspect.isclass(x) and issubclass(x, (Field, NXobject))
 
-        if isclass(name) or (isinstance(name, tuple) and len(name)
+        if isclass(name) or (isinstance(name, list) and len(name)
```
nvaytet, line 372: Should we consider having it accept both `list` and `tuple`?

nvaytet (approved):

PR #58, opened 2022-09-09
Fix exception when iterating NXobject that cannot be loaded

SimonHeybrock:
Without `__iter__` Python apparently just calls `__getitem__` with integers starting at 0 (until `IndexError`).

### DIFF
--- a/tests/nexus_test.py
+++ b/tests/nexus_test.py
@@ -58,7 +58,8 @@ def test_nxobject_items(nxroot):
 
 def test_nxobject_iter(nxroot):
     nxroot.create_class('entry2', NXentry)
-    # With missing __iter__ this used to raise since NXroot cannot be loaded.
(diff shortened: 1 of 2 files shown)


tests/nexus_test.py
```diff
@@ -56,6 +56,15 @@ def test_nxobject_items(nxroot):
             }
 
 
+def test_nxobject_iter(nxroot):
+    nxroot.create_class('entry2', NXentry)
+    # With missing __iter__ this used to raise since Python just calls __getitem__
```
nvaytet, line 61: Which is the part that used to raise? the `list(nxroot)` or the `for key in nxroot:`

SimonHeybrock: Both.

nvaytet (approved):

PR #62, opened 2022-10-17
Small fixes

SimonHeybrock:


docs/about/release-notes.rst
```diff
@@ -28,6 +28,15 @@ Release Notes
...
    and Jan-Lukas Wynen :sup:`a`
 
+v0.3.3 (October 2022)
+---------------------
+
+Bugfixes
+~~~~~~~~
+
+* Fix :meth:`scippnexus.Field.dims` and :meth:`scippnexus.Field.shape` to consistently return tuples instead of lists `#62 <https://github.com/scipp/scippnexus/pull/62>`_.
+* Fix :meth:`scippnexus.NXobject.__getitem__` to return children with correctly setup field dimensions when indexed with a class name `#62 <https://github.com/scipp/scippnexus/pull/62>`_.
```
jl-wynen, line 38:
```suggestion
* Fix :meth:`scippnexus.NXobject.__getitem__` to return children with correctly set up field dimensions when indexed with a class name `#62 <https://github.com/scipp/scippnexus/pull/62>`_.
```

src/scippnexus/nxobject.py
```diff
@@ -339,8 +338,9 @@ def _get_children_by_nx_class(
         children = {}
         select = tuple(select) if isinstance(select, list) else select
         for key in self.keys():
-            if issubclass(type(child := self._get_child(key)), select):
-                children[key] = child
+            if issubclass(type(self._get_child(key)), select):
+                # Get child again via __getitem__ so correct field dims are used.
+                children[key] = self[key]
```
jl-wynen, line 343: What was wrong with the old code?

SimonHeybrock: `_get_child` does not setup dims deduced from surrounding group, but we cannot use `__getitem__` in the loop as this leads to infinite recursion.

jl-wynen (approved):

PR #63, opened 2022-10-17
Application definitions

### LINKED ISSUE #65: Writing NeXus files based on application definitions
Follow-up to #63. Basically, given a data structure such as a `scipp.DataArray` and a NeXus application definition, we want to create/write groups. Currently `Nxobject` provides
https://github.com/scipp/scippnexus/blob/ece37dcceef024f31f2a08e19aecab3c48d07802/src/scippnexus/nxobject.py#L472

We can consider extending this with support for an application definition:

```python
with snx.File(name, definition=NXcanSAS) as f:
  group = f[path]
  group.create_class(name, definition=SASdata, data=my_data_array)
```
Here `definition` provides a key for lookup of the child strategy via `group._strategy` (which was setup from the `NXcanSAS` root definition). The child strategy must then provide everything necessary for writing the group and its content (attributes, fields+field attributes, child groups, ...) and how these relate to properties of the `data`.

Design wise, one key aspect to address is how to handle recursion. Should the method on NXobject be allowed to deal with this, i.e., the strategy may write an entire subtree, or should this be handled in another way, in a way that avoids handling the tree in the strategy?

An alternative opportunity to explore is whether `NXobject.__setitem__` could be generalized. Currently it only supports creations of *fields* (from `scipp.Variable`), as `scipp.DataArray` does not contain enough information to create a NeXus *group*. However, an application definition might provide a wrapper for this?

```python
group['sasdata01'] = SASdata(my_data_array)
```

Need to figure out if the dual purpose of `SASdata` as a definition/strategy for loading and a wrapper for data is a reasonable design.

SimonHeybrock:
Experimental (and incomplete) support for reading NXcanSAS.

Also fixes #65.

### DIFF
--- a/src/scippnexus/definitions/nxcansas.py
+++ b/src/scippnexus/definitions/nxcansas.py
@@ -4,15 +4,8 @@
 from typing import Optional
 from ..typing import H5Group
 
-# required children
-# required attrs
-# meaning of special attrs (e.g., errors)
-# consider structure verification out of scope, just deal with leaf class,
-# such as SASData!?
-
 
 def make_application_definition_strategy(application_definition, strategy):
-    print(application_definition, strategy)
 
     class ApplicationDefinitionStrategy(strategy):
         _application_definition = application_definition
--- a/src/scippnexus/definitions/nxcansas.py
+++ b/src/scippnexus/definitions/nxcansas.py
@@ -14,7 +14,7 @@ from ..typing import H5Group
 class Definition:
 
     @classmethod
-    def definition_for_group(cls, group: H5Group):
+    def make_strategy(cls, group: H5Group):
         if (definition_class := group.attrs.get(cls.class_attribute)) is not None:
             return cls.definitions.get(definition_class)
 
--- a/src/scippnexus/definitions/nxcansas.py
+++ b/src/scippnexus/definitions/nxcansas.py
@@ -3,22 +3,6 @@
 # @author Simon Heybrock
 import scipp as sc
 from typing import Optional
-from ..typing import H5Group
-
-
-def make_application_definition_strategy(application_definition, strategy):
-
-    class ApplicationDefinitionStrategy(strategy):
-
-        @staticmethod
-        def __class_attribute__():
-            return application_definition._class_attribute
-
-        @staticmethod
-        def child_strategy(group: H5Group):
-            return application_definition.child_strategy(group)
-
-    return ApplicationDefinitionStrategy
 
 
 class ApplicationDefinition:
@@ -28,7 +12,7 @@ class ApplicationDefinition:
         self._class_attribute = class_attribute
         self._strategies = {}
 
-    def child_strategy(self, group):
+    def make_strategy(self, group):
         # This approach will likely need to be generalized as many application
         # definitions to not define a "class attribute" in the style of canSAS_class,
         # but seem to rely on basic strcture and the NX_class attribute.
@@ -36,16 +20,15 @@ class ApplicationDefinition:
                                                 self._default_class)) is not None:
             return self._strategies.get(definition_class)
 
-    def __call__(self, strategy):
-        strat = make_application_definition_strategy(self, strategy)
-        self._strategies[strategy.__name__] = strat
-        return strat
+    def register(self, strategy):
+        self._strategies[strategy.__name__] = strategy
+        return strategy
 
 
 NXcanSAS = ApplicationDefinition('canSAS_class', 'SASroot')
 
 
-@NXcanSAS
+@NXcanSAS.register
 class SASdata:
     nx_class = 'NXdata'
 
@@ -114,7 +97,7 @@ class SASdata:
         raise RuntimeError("Cannot handle both uncertainties and resolutions for Q")
 
 
-@NXcanSAS
+@NXcanSAS.register
 class SAStransmission_spectrum:
 
     @staticmethod
@@ -125,7 +108,7 @@ class SAStransmission_spectrum:
         return ('lambda', )
 
 
-@NXcanSAS
+@NXcanSAS.register
 class SASentry:
     nx_class = 'NXentry'
 
@@ -141,6 +124,6 @@ class SASentry:
         group.create_field('run', self.run)
 
 
-@NXcanSAS
+@NXcanSAS.register
 class SASroot:
     pass
--- a/src/scippnexus/definitions/nxcansas.py
+++ b/src/scippnexus/definitions/nxcansas.py
@@ -11,21 +11,45 @@ from ..typing import H5Group
 # such as SASData!?
 
 
-class ApplicationDefinitionStrategy:
+def make_application_definition_strategy(application_definition, strategy):
+    print(application_definition, strategy)
 
-    def __init__(self, group: H5Group):
-        self._group = group
+    class ApplicationDefinitionStrategy(strategy):
+        _application_definition = application_definition
 
-    def child_strategy(self, group: H5Group):
-        if (definition_class := group.attrs.get(self.class_attribute)) is not None:
-            return self.strategies.get(definition_class)
+        def __init__(self, group: H5Group):
+            self._group = group
+            super().__init__()
+
+        def child_strategy(self, group: H5Group):
+            return self._application_definition.child_strategy(group)
+
+    return ApplicationDefinitionStrategy
 
 
-class NXcanSAS(ApplicationDefinitionStrategy):
-    class_attribute = 'canSAS_class'
+class ApplicationDefinition:
+
+    def __init__(self, class_attribute, default=None):
+        self._default_class = default
+        self._class_attribute = class_attribute
+        self._strategies = {}
+
+    def child_strategy(self, group):
+        if (definition_class := group.attrs.get(self._class_attribute,
+                                                self._default_class)) is not None:
+            return self._strategies.get(definition_class)
+
+    def __call__(self, strategy):
+        strat = make_application_definition_strategy(self, strategy)
+        self._strategies[strategy.__name__] = strat
+        return strat
 
 
-class SASdata(NXcanSAS):
+NXcanSAS = ApplicationDefinition('canSAS_class', 'SASroot')
+
+
+@NXcanSAS
+class SASdata:
 
     @property
     def dims(self):
@@ -72,7 +96,8 @@ class SASdata(NXcanSAS):
         raise RuntimeError("Cannot handle both uncertainties and resolutions for Q")
 
 
-class SAStransmission_spectrum(NXcanSAS):
+@NXcanSAS
+class SAStransmission_spectrum:
 
     @property
     def dims(self):
@@ -82,13 +107,11 @@ class SAStransmission_spectrum(NXcanSAS):
         return ('lambda', )
 
 
-class SASentry(NXcanSAS):
+@NXcanSAS
+class SASentry:
     pass
 
 
-class SASroot(NXcanSAS):
+@NXcanSAS
+class SASroot:
     pass
-
-
-NXcanSAS.classes = [SASroot, SASentry, SASdata, SAStransmission_spectrum]
-NXcanSAS.strategies = {cls.__name__: cls for cls in NXcanSAS.classes}
--- a/src/scippnexus/definitions/nxcansas.py
+++ b/src/scippnexus/definitions/nxcansas.py
@@ -38,10 +38,10 @@ class SASdata:
 
     def __init__(self,
                  data: sc.DataArray,
-                 variances: Optional[Literal['uncertainties', 'resolutions']] = None):
+                 Q_variances: Optional[Literal['uncertainties', 'resolutions']] = None):
         self.data = data
-        assert variances in (None, 'uncertainties', 'resolutions')
-        self._variances = variances
+        assert Q_variances in (None, 'uncertainties', 'resolutions')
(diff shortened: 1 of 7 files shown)


src/scippnexus/definitions/nxcansas.py
```diff
@@ -0,0 +1,94 @@
...
+# such as SASData!?
+
+
+class ApplicationDefinitionStrategy:
+
+    def __init__(self, group: H5Group):
+        self._group = group
+
+    def child_strategy(self, group: H5Group):
+        if (definition_class := group.attrs.get(self.class_attribute)) is not None:
```
jl-wynen, line 20: This requires child classes to define `class_attribute` and `strategies`. @nvaytet and I talked about this recently and concluded that we should avoid this kind of requirement.
Can we avoid it here? That is, can this be implemented by defining those fields in `ApplicationDefinitionStrategy`? Or can we use a decorator instead of inheritance?

src/scippnexus/definitions/nxcansas.py
```diff
@@ -0,0 +1,94 @@
...
+            return self.strategies.get(definition_class)
```
jl-wynen, line 21: This function returns `None` when 
- there is no class attribute in the file
- there is no child strategy for the given attribute

Shouldn't the 2nd case be a hard failure, i.e. something is not supported by scippnexus?

src/scippnexus/definitions/nxcansas.py
```diff
@@ -0,0 +1,94 @@
...
+
+class SASentry(NXcanSAS):
+    pass
+
+
+class SASroot(NXcanSAS):
+    pass
+
+
+NXcanSAS.classes = [SASroot, SASentry, SASdata, SAStransmission_spectrum]
```
jl-wynen, line 93: By inheriting from `NXcanSAS` in all classes, this application definition supports arbitrary nesting. E.g. a root inside data. Does this make sense? Should we limit it more? For instance using something like the nexus loader prototype in ess?

src/scippnexus/nxdata.py
```diff
@@ -2,7 +2,7 @@
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
```
jl-wynen, line 2: This approach seems quite invasive in that it requires the `NX*` classes to know about and handle strategies. Is this necessary or could it be implemented on a higher level like the loader prototype in ess?
So what kind of modification to the behaviour of, e.g. `NXdata` does the application definition need to make which cannot be done as post processing?

jl-wynen:
Overall comment to the approach in https://github.com/scipp/scippnexus/pull/63#discussion_r997889644

src/scippnexus/definitions/nxcansas.py
SimonHeybrock: I think that is for the application definition to decide. Without a custom strategy, child strategies are always set to `None`, meaning the default strategy will be used by the child. Conceptually, this means the parent "does not care about or control" the child.

src/scippnexus/definitions/nxcansas.py
SimonHeybrock: I had considered this. My conclusion was that in many cases structure may not be something people want to enforce (it will prevent reading partially broken/non-standard files, for example). At the same time, you *can* write your strategy in a way that it enforces structure.

src/scippnexus/nxdata.py
SimonHeybrock: > This approach seems quite invasive in that it requires the NX* classes to know about and handle strategies.

Well, this is the strategy pattern. It does not have to know about the details of the strategy.

> Is this necessary or could it be implemented on a higher level like the loader prototype in ess?

Not really, you would need to fall back to `h5py`, since `scippnexus` will generally not be able to load groups that follow an application definition, e.g., because they define axes (dims) in a custom manner.

> So what kind of modification to the behaviour of, e.g. NXdata does the application definition need to make which cannot be done as post processing?

Probably some could be made. But the post processing would need to reopen the file and read custom attributes, etc. So basically you could just go back and implemented your loader with plain h5py instead of scippnexus, as there might be little benefit.

src/scippnexus/definitions/nxcansas.py
SimonHeybrock: Sure, this file is just a quick draft, to see if the strategy mechanism can be used for this. There are many things to improve here.

SimonHeybrock:
@jl-wynen I refactored to use decorators (I will move this to a separate file once we are happy with it). One thing that bothers me is the creation of instances of the strategies for each group. I have considering changing the `self` argument of, e.g., `NXdataStrategy` or `SASdata` to `group`, and remove the `__init__`s that store the group. Then, instead of instantiating in the Nexus class, we would just call `self._strategy.dims(self)` instead of `self._strategy.dims`. This also looks a bit odd, but the strategy implementations themselves would be cleaner. Not they always have `self._group.bla`, with `_group` defined invisibly in a base class via the decorator.

SimonHeybrock:
> @jl-wynen I refactored to use decorators (I will move this to a separate file once we are happy with it). One thing that bothers me is the creation of instances of the strategies for each group. I have considering changing the `self` argument of, e.g., `NXdataStrategy` or `SASdata` to `group`, and remove the `__init__`s that store the group. Then, instead of instantiating in the Nexus class, we would just call `self._strategy.dims(self)` instead of `self._strategy.dims`. This also looks a bit odd, but the strategy implementations themselves would be cleaner. Note they always have `self._group.bla`, with `_group` defined invisibly in a base class via the decorator.

I made this change, since it will make support for writing more convenient. Not perfectly happy with the result, but not worse that before either.

SimonHeybrock:
@jl-wynen I managed to cleanup the mechanism significantly. Can you have a look?

src/scippnexus/nxdata.py
```diff
@@ -11,35 +11,72 @@
 from .nxobject import Field, NXobject, ScippIndex, NexusStructureError, asarray
 
 
+class NXdataStrategy:
```
jl-wynen, line 14: How would a user define a new strategy?
`SASdataStrategy` does not seem to be fully in line with `NXdataStrategy` as it has an additional public method. So is there any code that needs to know the concrete strategy type?

src/scippnexus/definitions/nxcansas.py
```diff
@@ -0,0 +1,141 @@
...
+            # to me what the difference is.
+            coord.attrs['resolutions'] = 'Qdev'
+            group.create_field('Qdev', sc.stddevs(da.coords['Q']))
+
+
+@NXcanSAS.register('SASdata')
+class SASdataStrategy:
+
+    @staticmethod
+    def dims(group: NXobject) -> Tuple[str]:
```
jl-wynen, line 64: What is is used for? It is not part of the 'base' strategy.

tests/application_definition_test.py
```diff
@@ -0,0 +1,38 @@
...
+import scipp as sc
+from scippnexus import NXroot, NXentry
+from scippnexus.definitions.nxcansas import NXcanSAS, SASentry, SASdata
+import pytest
+
+
+@pytest.fixture()
+def nxroot(request):
+    """Yield NXroot containing a single NXentry named 'entry'"""
+    with h5py.File('dummy.nxs', mode='w', driver="core", backing_store=False) as f:
```
jl-wynen, line 11: Keep this file in memory instead of going out to disk.

src/scippnexus/definitions/nxcansas.py
```diff
@@ -0,0 +1,141 @@
...
+NXcanSAS = ApplicationDefinition('canSAS_class', 'SASroot')
+
+
+class SASdata:
+    nx_class = 'NXdata'
+
+    def __init__(self, data):
+        self.data = data
+
+    def __write_to_nexus_group__(self, group: NXobject):
```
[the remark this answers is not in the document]
SimonHeybrock: The reason for not making this a normal method was that users may want to make their existing classes "writable", without implementing a new class, and without changing the publicly visible interface.

src/scippnexus/definitions/nxcansas.py
```diff
@@ -0,0 +1,141 @@
...
+        signal = group.create_field('I', sc.values(da.data))
+        if da.variances is not None:
+            signal.attrs['uncertainties'] = 'Idev'
+            group.create_field('Idev', sc.stddevs(da.data))
+        coord = group.create_field('Q', da.coords['Q'])
+        if da.coords['Q'].variances is not None:
+            # Note that there is also an "uncertainties" attribute. It is not clear
+            # to me what the difference is.
+            coord.attrs['resolutions'] = 'Qdev'
+            group.create_field('Qdev', sc.stddevs(da.coords['Q']))
```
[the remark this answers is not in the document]
SimonHeybrock: Yes, if someone actually uses both they will need to provide this in a different manner.

src/scippnexus/definitions/nxcansas.py
```diff
@@ -0,0 +1,141 @@
...
+@NXcanSAS.register('SASdata')
+class SASdataStrategy:
+
+    @staticmethod
+    def dims(group: NXobject) -> Tuple[str]:
+        axes = group.attrs.get('axes')
+        if isinstance(axes, str):
+            axes = axes.split(" ")
+        if axes is None:
+            axes = group.attrs.get('I_axes')
```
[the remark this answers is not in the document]
SimonHeybrock: The documentation does not say. Will probably need to fix this in the future if it causes problems.

src/scippnexus/definitions/nxcansas.py
```diff
@@ -0,0 +1,141 @@
...
+        if not isinstance(axes, list):
```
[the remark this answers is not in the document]
SimonHeybrock: 👍 Checking for `int` now instead.

src/scippnexus/nxobject.py
```diff
@@ -138,7 +138,9 @@ class Field:
     In HDF5 fields are represented as dataset.
     """
 
-    def __init__(self, dataset: H5Dataset, dims=None, is_time=None):
+    def __init__(self, dataset: H5Dataset, *, parent, dims=None, is_time=None):
+        # TODO May be grandparent in some cases
```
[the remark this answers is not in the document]
SimonHeybrock: I think it does not, I will remove this comment, and rename to `ancestor`?

src/scippnexus/nxdata.py
SimonHeybrock: Hmm, I suppose you refer to `dims`? It looks unused and is probably leftover due to an oversight during a refactor.

SimonHeybrock: I remember now: It was from a time when I tried to support broken files (not following NXcanSAS), I will remove.

src/scippnexus/definitions/nxcansas.py
SimonHeybrock: See above.

tests/application_definition_test.py
SimonHeybrock: Isn't that what `backing_store=False` does?

src/scippnexus/definitions/nxcansas.py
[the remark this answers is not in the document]
jl-wynen: The dunder method is public. In my mind, a dunder method relates to a general convention across libraries / codebases. In this case here, we basically have duck typing, we just require a certain method.

src/scippnexus/definitions/nxcansas.py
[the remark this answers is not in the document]
jl-wynen: But shouldn't we write uncertainties instead of resolutions?

tests/application_definition_test.py
jl-wynen: Ah yes, I didn't know this option.

src/scippnexus/definitions/nxcansas.py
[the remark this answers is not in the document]
SimonHeybrock: Yes public, but it does not interfere with tab completion, etc.

src/scippnexus/definitions/nxcansas.py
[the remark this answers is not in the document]
SimonHeybrock: Usually people talk about "Q-resolution", and I do not think it is statistical. I don't know if anyone has decided how it will be stored, and for now just assumed they might use Scipp's variances.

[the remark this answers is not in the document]
SimonHeybrock: We will definitely need to check with the scientists, for now a lot of this is incomplete anyway.

src/scippnexus/definitions/nxcansas.py
```diff
@@ -36,8 +36,12 @@ def decorator(strategy):
 class SASdata:
     nx_class = 'NXdata'
 
-    def __init__(self, data):
+    def __init__(self,
+                 data: sc.DataArray,
+                 Q_variances: Optional[Literal['uncertainties', 'resolutions']] = None):
         self.data = data
+        assert Q_variances in (None, 'uncertainties', 'resolutions')
```
jl-wynen, line 43: I learned a surprising thing recently, assertions are removed when making optimised builds (e.g. in a wheel). So this should be a proper `if-raise` construct.

SimonHeybrock: Didn't know that either!

PR #74, opened 2022-11-07
Document application definition mechanism

SimonHeybrock:
Also properly expose and define interfaces.


src/scippnexus/definition.py
```diff
@@ -0,0 +1,14 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+# @author Simon Heybrock
+from abc import ABC, abstractmethod
+from .nxobject import NXobject
+
+
+class ApplicationDefinition(ABC):
```
jl-wynen, line 8: Add to the docs?

src/scippnexus/definitions/nxcansas.py
```diff
@@ -4,9 +4,10 @@
 import scipp as sc
 from typing import Optional, Tuple, Union, Literal
 from ..nxobject import NXobject
+from ..definition import ApplicationDefinition as BaseDef
 
 
-class ApplicationDefinition:
+class ApplicationDefinition(BaseDef):
```
jl-wynen, line 10: Is it intentional that NXCanSAS is not included in the docs?

docs/user-guide/application-definitions.ipynb
```diff
@@ -0,0 +1,211 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f1dbfbe6-9605-4f1a-b8c9-a31768df60f5",
+   "metadata": {},
+   "source": [
+    "## Overview\n",
+    "\n",
+    "[NeXus Application Definitions](https://manual.nexusformat.org/classes/applications/index.html) define mandatory and optional class contents for specific applications.\n",
+    "ScippNexus' approach to application definitions is to consider them as a guide, without performing full validation.\n",
```
jl-wynen, line 19: Do we write it as 'ScippNexus' or 'ScippNeXus'?

docs/user-guide/application-definitions.ipynb
```diff
@@ -0,0 +1,211 @@
...
+    "This is to avoid getting in the way of the library user, e.g., when working with incomplete or partially broken files.\n",
+    "For example, ScippNexus will generally not validate that the tree structure of a file following an application definition is conforming."
```
jl-wynen, line 21:
```suggestion
    "For example, ScippNexus will generally not validate that the tree structure conforms to a given application definition."
```
The original reads a bit strangely.

docs/user-guide/application-definitions.ipynb
```diff
@@ -0,0 +1,211 @@
...
+    "\n",
+    "\n",
+    "class MyData:\n",
+    "    nx_class = snx.NXdata  # required\n",
+    "\n",
+    "    def __init__(self, data: sc.DataArray):\n",
+    "        self._data = data\n",
+    "\n",
+    "    def __write_to_nexus_group__(self, group: snx.NXobject):  # required\n",
+    "        group.attrs['axes'] = self._data.dims  # NeXus way of defining dim labels\n",
```
jl-wynen, line 74:
```suggestion
    "        # NeXus way of defining dim labels\n",
    "        group.attrs['axes'] = self._data.dims\n",
```
The line is too long

docs/user-guide/application-definitions.ipynb
```diff
@@ -0,0 +1,211 @@
...
+    "        group['mysignal'] = self._data.data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "11c5073d-7231-429c-a461-61b31a03c284",
+   "metadata": {},
+   "source": [
+    "Note that above we use a custom \"signal\" name and fail to set the \"signal\" attribute on the group and as such deviate from the NeXus specification.\n",
```
jl-wynen, line 83:
```suggestion
    "Note that above we use a custom \"signal\" name and do not to set the \"signal\" attribute on the group and as such deviate from the NeXus specification.\n",
```
The original sounds like the function is broken.

docs/user-guide/application-definitions.ipynb
```diff
@@ -0,0 +1,211 @@
...
+    "## Reading files\n",
+    "\n",
+    "For some application definitions &mdash; or classes within application definitions &mdash; the default ScippNexus mechanisms for reading are sufficient.\n",
+    "This is the case when the application definition follows the NeXus standard and, e.g., introduces no new attributes.\n",
+    "\n",
+    "In other cases we require customization of how ScippNexus reads class contents.\n",
+    "On the lowest level, this is handled using *strategies* that can be passed to `scippnexus.NXobject` and its subclasses such as `scippnexus.NXdata`.\n",
+    "Handling of strategies is usually encapsulated by a *definition* that can be passed to `scippnexus.File`.\n",
+    "\n",
+    "As an example, consider the following simple strategy for loading data with a custom signal name, which the file failed to specify.\n",
```
jl-wynen, line 116:
```suggestion
    "As an example, consider the following simple strategy for loading data with a custom signal name, which the file does not specify.\n",
```
Again, it sounds like the file is broken and by extensions that application definitions are about handling broken files.

src/scippnexus/__init__.py
```diff
@@ -14,3 +14,8 @@
 from .nxobject import NXobject, Field, Attrs
 from .nxobject import NexusStructureError
 from .nexus_classes import *
+
+from .nxdata import NXdataStrategy
+from .nxdetector import NXdetectorStrategy
+from .nxlog import NXlogStrategy
+from .definition import ApplicationDefinition
```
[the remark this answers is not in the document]
SimonHeybrock: Don't see good reason why.

src/scippnexus/definitions/nxcansas.py
SimonHeybrock: Yes, it is incomplete and I am not sure it will stay, and it is not necessary for the other docs.

docs/user-guide/application-definitions.ipynb
SimonHeybrock: The former. ScippNeXus looks bad and is a nightmare to type :D

docs/user-guide/application-definitions.ipynb
SimonHeybrock: Yes, the file is broken, it is not NeXus. For example, NXcanSAS is not NeXus.

src/scippnexus/definition.py
SimonHeybrock: Done

docs/user-guide/application-definitions.ipynb
jl-wynen: It is ok for me if I have the window on the full screen or collapse the sidebars. But it is really difficult to tune this for everyone. We should just make sure that we don't use our large monitors for testing.

docs/user-guide/application-definitions.ipynb
jl-wynen: This is maybe a philosophical questions. But I would see application definitions as part of NeXus. So if a file+app.def. can be loaded, it is valid NeXus.

But my point is more about the impression we give to users. To me, it sounds like app.defs. are about working around broken files, where in reality they are about providing a schema.

jl-wynen (approved):

docs/user-guide/application-definitions.ipynb
SimonHeybrock: Good point, considering file+app.def as a unit.

I would have liked to give a different example, but right now all supported customizations points in the strategies revolve around making things that are not plain NeXus work.

PR #88, opened 2022-12-21
Support loading fields of NXdata or NXdetector with no signal

SimonHeybrock:
This does not mean we can load the group as a whole, but at least we avoid errors when trying to load fields within the group.

### DIFF
--- a/src/scippnexus/nxdata.py
+++ b/src/scippnexus/nxdata.py
@@ -141,14 +141,14 @@ class NXdata(NXobject):
 
         Does not check for potential bin-edge coord.
         """
+        shape = self._get_child(name).shape
+        if self.shape == shape:
+            return self.dims
         lut = {}
         if self._signal is not None:
             for d, s in zip(self.dims, self.shape):
                 if self.shape.count(s) == 1:
                     lut[s] = d
-        shape = self._get_child(name).shape
-        if self.shape == shape:
-            return self.dims
         try:
             dims = [lut[s] for s in shape]
         except KeyError:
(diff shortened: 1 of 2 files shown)


src/scippnexus/nxdata.py
```diff
@@ -139,9 +142,10 @@ def _guess_dims(self, name: str):
         Does not check for potential bin-edge coord.
         """
         lut = {}
-        for d, s in zip(self.dims, self.shape):
-            if self.shape.count(s) == 1:
-                lut[s] = d
+        if self._signal is not None:
+            for d, s in zip(self.dims, self.shape):
+                if self.shape.count(s) == 1:
+                    lut[s] = d
```
jl-wynen, line 148: You could move this entire block that construct `lut` below `return self.dim`. And then use
```python
if self._signal is None:
    raise NexusStructureError(...)
```

SimonHeybrock: ... but we don't want to raise?

SimonHeybrock: Oh, wait, it is wrapped where it catches.

SimonHeybrock: I moved the block, but I cannot see how I can raise unconditionally.

jl-wynen: It already raises unconditionally. If signal is None, `lut` is empty and the `try-except` block below will get a KeyError and raise.

Unless `shape` can be empty. Can this be the case?

SimonHeybrock: Yes, `shape` can be empty.

jl-wynen: ok

jl-wynen (approved):

PR #92, opened 2023-01-11
Try pre-commit-ci

jl-wynen:


jl-wynen:
The latency of pre-commit-ci is too high. The test suite runs almost in full before pre-commit-ci's auto fix. So I will not pursue this approach further for now.

PR #93, opened 2023-01-11
Pre commit in ci

jl-wynen:


tox.ini
```diff
@@ -35,13 +35,11 @@ commands = python -m sphinx -j2 -v -b linkcheck -d {toxworkdir}/docs_doctrees do
 
 [testenv:static]
 description = Code formatting and static analysis
-basepython = python3.8
+basepython = python3.10
 skip_install = true
 deps = -r requirements/static.txt
 allowlist_externals=bash
```
SimonHeybrock, line 41: Can remove this line?

jl-wynen:
I cleaned up the commits to avoid clutter. But this setup seems to work. It runs all pre-commit checks. If static analysis fails, it stops the workflow. If files get reformatted, it commits the result. And as before, CI keeps running with the old files and never with the autocommited ones.

tox.ini
```diff
@@ -35,13 +35,10 @@ commands = python -m sphinx -j2 -v -b linkcheck -d {toxworkdir}/docs_doctrees do
...
 description = Code formatting and static analysis
-basepython = python3.8
 skip_install = true
 deps = -r requirements/static.txt
-allowlist_externals=bash
-commands = nbstripout --extra-keys 'metadata.language_info.version cell.metadata.jp-MarkdownHeadingCollapsed' --drop-empty-cells .
-           yapf --recursive --in-place src tests
-           flake8 .
+commands = - pre-commit run -a
+           pre-commit run -a
```
SimonHeybrock, line 41: Can you add a comment explaining what this does?

jl-wynen: Like this?

jl-wynen:
Note for future reference, that it would be nicer to run it like this in a shell `pre-commit run -a || pre-commit run -a`, because it only runs the 2nd command if the 1st fails. But tox does not seem to support this.

tox.ini
SimonHeybrock: Could it be written so it does the second one only if the first one fails? Something like `pre-commit run -a || pre-commit run -a` maybe?

jl-wynen: https://github.com/scipp/scippnexus/pull/93#issuecomment-1381545490

jl-wynen: It would need an extra script file, I think.

SimonHeybrock:
Why not something like
```
commands =
  sh -c 'pre-commit run -a || pre-commit run -a'
```

SimonHeybrock (approved):

jl-wynen:
Works. I tried something like this before and it gave a syntax error in tox. I don't remember what I did differently.

I also changed the order of checks in `.pre-commit-config.yaml` such that formatters run before analysers. This means, e.g., that flake8 does not report on formatting issues that yapf can fix.

SimonHeybrock:
> I also changed the order of checks in .pre-commit-config.yaml such that formatters run before analysers. This means, e.g., that flake8 does not report on formatting issues that yapf can fix.

Can you make the same change in scipp please? Thanks!

jl-wynen:
> Can you make the same change in scipp please? Thanks!

Yes. I am going to roll this out everywhere.

PR #94, opened 2023-01-17
Do not create event weights with variances

### LINKED ISSUE #91: Should loading event data no longer create weights with variances?
After scipp/scipp#2895, broadcasting of variances will no longer be supported. This implies that the "standard" paradigm of handling uncertainties will not be feasible any more. This further implies that there is limited use to carrying uncertainties of "counts", since uncertainties can simply be computed later on.

We should therefore consider avoiding the overhead of creating variances for the weights when an `NXdetector` with `NXevent_data` is loaded. This would save both memory and compute resources.

SimonHeybrock:
Fixes #91.

### DIFF
--- a/docs/about/release-notes.rst
+++ b/docs/about/release-notes.rst
@@ -34,7 +34,7 @@ v23.01.0
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
-* :class:`scippnexus.NXevent_data` is not loaded without automatically assigning a variance of 1 to the even weights `#92 <https://github.com/scipp/scippnexus/pull/92>`_.
+* :class:`scippnexus.NXevent_data` is not loaded without automatically assigning a variance of 1 to the even weights `#94 <https://github.com/scipp/scippnexus/pull/94>`_.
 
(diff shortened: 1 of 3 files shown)


docs/about/release-notes.rst
```diff
@@ -28,6 +28,19 @@ Release Notes
    Neil Vaytet :sup:`a`\ ,
    and Jan-Lukas Wynen :sup:`a`
 
+v23.01.0
+--------
+
+Breaking changes
+~~~~~~~~~~~~~~~~
+
+* :class:`scippnexus.NXevent_data` is not loaded without automatically assigning a variance of 1 to the even weights `#94 <https://github.com/scipp/scippnexus/pull/94>`_.
```
jl-wynen, line 37:
```suggestion
* :class:`scippnexus.NXevent_data` is now loaded without automatically assigning a variance of 1 to the event weights `#94 <https://github.com/scipp/scippnexus/pull/94>`_.
```

jl-wynen (approved):

PR #95, opened 2023-01-19
Support nested loading as DataGroup and change error handling

### LINKED ISSUE #73: Change `nexus.NXobject` interface to use exceptions rather than warnings
`load_nexus` relies on warnings if it cannot load certain parts of the file, to ensure that something incomplete rather than nothing is returned to the users.

For the new lower-level interface around `NXobject` this does not seem like the right choice.

 - Refactor to move warning-generating code into functions used by `load_nexus`, use exceptions in low-level functions shared with `NXobject` subclasses.

### LINKED ISSUE #85: (Nested) Loading of groups
In Scipp, we are considering the addition of a `DataGroup` container. This would be similar to `Dataset`, but without coords and without restricting the dims or shapes of the items. This is thus quite similar to a Nexus "group". We would therefore like to support loading groups in ScippNexus, returning a `DataGroup`. There are a number of things to consider:

- #57 changed `__getitem__` to return Python scalars instead of `scipp.Variable` if not shape are unit is given. This was for more convenient storage in a Python `dict`. For `DataGroup`, the we are currently leaning towards requiring items to have `dims` and `shape`. Should we thus undo this change in ScippNexus? Or should `DataGroup` be more flexible?
- How to handle errors while loading? In practice these are unfortunately quite frequent, so failing the entire load may not be so useful. One promising approach would be to fall back to a "plain" load as (nested) `DataGroup`, since most errors are from "higher level" logic, such as trying to interpret fields for an NXevent_data or NXdetector group. There are a number of subtleties here, especially implementation wise, as the current design puts some hurdles here.
  - One subproblem is the handling of coords that fail to insert due to a `DimensionError`. Currently these are skipped with a warning. We could instead return the entire `NXdata` as `DataGroup`, but this would likely not be useful in many cases. But not doing that would be inconsistent.
- Handling of links and references. For example, NXevent_data (or its fields) has two places that reference it in SNS files. I think any attempt to do something "smart" like returning a tree with sharing setup will just cause headaches. I feel we should probably make a clear statement to not support this (i.e., data will simply be loaded twice).

SimonHeybrock:
Fixes #85.
Fixes #73.

There are two major changes here:

1. Any NXobject can now be loaded as `scipp.DataGroup`.
2. If an implementation of NXobject fails to load (e.g., as a `DataArray`), the implementation will fall back to a generic load of all datasets in subgroups as a `DataGroup`. Previously, this was handled in a variety of cases by simply skipping the offending dataset or group.

Note that this depends on scipp 23.01.0, and I will need to run `pip-compile-multi` after that is released. For now the builds will fail.

While 2.) is generally an improvement, there are cases where it hurts usability. For example, for a WISH file we previously got:
```python
/home/simon/code/scipp/scippnexus/src/scippnexus/nxdata.py:243: UserWarning: Skipped load of axis /raw_data_1/instrument/detector_1/event_frame_number due to:
Expected dimension to be in [spectrum_index:97280, ], got dim_0.
  warn(f"Skipped load of axis {field.name} due to:\n{e}")
/home/simon/code/scipp/scippnexus/src/scippnexus/nxdata.py:243: UserWarning: Skipped load of axis /raw_data_1/instrument/detector_1/event_time_bins due to:
Expected dimension to be in [spectrum_index:97280, ], got dim_0.
  warn(f"Skipped load of axis {field.name} due to:\n{e}")
/home/simon/code/scipp/scippnexus/src/scippnexus/nxdata.py:243: UserWarning: Skipped load of axis /raw_data_1/instrument/detector_1/time_of_flight due to:
Expected dimension to be in [spectrum_index:97280, ], got dim_0.
  warn(f"Skipped load of axis {field.name} due to:\n{e}")
/home/simon/code/scipp/scippnexus/src/scippnexus/nxobject.py:295: UserWarning: Unrecognized unit 'pulses' for value dataset in '/raw_data_1/instrument/detector_1/time_of_flight_raw'; setting unit as 'dimensionless'
  warnings.warn(f"Unrecognized unit '{unit}' for value dataset "
/home/simon/code/scipp/scippnexus/src/scippnexus/nxdata.py:243: UserWarning: Skipped load of axis /raw_data_1/instrument/detector_1/time_of_flight_raw due to:
Expected dimension to be in [spectrum_index:97280, ], got dim_0.
  warn(f"Skipped load of axis {field.name} due to:\n{e}")
```
with a (partially) successful load as data binned into pixels:

image

Now we get this instead:

image

This is more complete, but less useful for the average user.


src/scippnexus/nxtransformations.py
```diff
@@ -65,7 +65,10 @@ def __getitem__(self, select: ScippIndex):
         # According to private communication with Tobias Richter, NeXus allows 0-D or
         # shape=[1] for single values. It is unclear how and if this could be
         # distinguished from a scan of length 1.
-        t = self._obj[select] * self.vector
+        value = self._obj[select]
+        if isinstance(value, sc.DataGroup):
+            raise TransformationError(f"Failed to load transformation at {self.name}.")
```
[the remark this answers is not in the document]
SimonHeybrock: Good question. I am having a closer look. I might need some additional changes.

tests/nxtransformations_test.py
```diff
@@ -178,3 +178,31 @@ def test_chain_with_multiple_values_and_different_time_unit(nxroot):
     expected = sc.spatial.affine_transform(value=np.identity(4), unit=t.unit)
     expected = t * (t * (offset * expected))
     assert sc.identical(detector[...].coords['depends_on'].value, expected)
+
+
+def test_broken_time_dependent_transformation_returns_path(nxroot):
```
jl-wynen: Can you indicate why this transformation is broken?

src/scippnexus/nxobject.py
```diff
@@ -382,11 +382,24 @@ def _get_child(
...
             da = NXobject._getitem(self, name)
         if (t := self.depends_on) is not None:
-            if hasattr(da, 'coords'):
-                da.coords['depends_on'] = t if isinstance(t,
-                                                          sc.Variable) else sc.scalar(t)
-            else:
-                da['depends_on'] = t
+
+            def insert(name, obj):
+                if hasattr(da, 'coords'):
```
jl-wynen, line 387: What types can `da` be? Is this check equivalent to `if not isinstance(da, sc.DataGroup)`?

SimonHeybrock: I think it is actually only `DataGroup` or `DataArray`. At some point we loaded some things as `Dataset`.

tests/nxtransformations_test.py
```diff
@@ -178,3 +178,31 @@ def test_chain_with_multiple_values_and_different_time_unit(nxroot):
...
+        sc.array(dims=['time'], values=[1.1, 2.2], unit='m'),
+        coords={'time': sc.array(dims=['time'], values=[11, 22], unit='s')})
+    log.coords['time'] = sc.epoch(unit='ns') + log.coords['time'].to(unit='ns')
+    offset = sc.spatial.translation(value=[1, 2, 3], unit='m')
+    vector = sc.vector(value=[0, 0, 1])
+    t = log * vector
+    t.data = sc.spatial.translations(dims=t.dims, values=t.values, unit=t.unit)
+    value = transformations.create_class('t1', NXlog)
+    value['time'] = log.coords['time'] - sc.epoch(unit='ns')
+    value['value'] = log.data[0:0]
```
SimonHeybrock, line 197:
```suggestion
    # This makes the transform "broken" since "time" has length 2 but data has length 0.
    value['value'] = log.data[0:0]
```

SimonHeybrock, line 197:
```suggestion
    # This makes the transform "broken" since "time" has length 2 but data has length 0.
    value['value'] = log.data[0:0]
```

SimonHeybrock:
@jl-wynen Do you have more comments on this? Builds are passing now, I think.

jl-wynen (approved):

PR #97, opened 2023-01-30
Catch more errors from NXtransformations

SimonHeybrock:
Will need to make a bugfix release, since this broke scippneutron.


jl-wynen (approved):
To help with these things in the future, can you make a page in the docs that explains what exceptions can be raised and what they mean? In particular, which exceptions trigger fallback loaders or other special behaviour.

PR #98, opened 2023-02-02
Ignore cue_* fields in choppers which interfere with loading

SimonHeybrock:
Similar to the same mechanism we use for detector data.

### DIFF
--- a/src/scippnexus/nxlog.py
+++ b/src/scippnexus/nxlog.py
@@ -58,7 +58,9 @@ class NXlog(NXobject):
 
     @property
     def _nxbase(self) -> NXdata:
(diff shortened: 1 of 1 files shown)


SimonHeybrock:
They provide an index into a potentially long dimension, from the Nexus docs:

```
In order to make random access to timestamped data faster there is an optional array pair of cue_timestamp_zero and cue_index. The cue_timestamp_zero will contain coarser timestamps than in the time array, say every five minutes. The cue_index will then contain the index into the time,value pair of arrays for that coarser cue_timestamp_zero.
```

These fields have a different length than the data, so they prevent loading it as a `DataArray`. Since they contain no actual information we skip them.

jl-wynen (approved):

PR #99, opened 2023-02-02
Handle issues with getting field dims more gracefully

SimonHeybrock:
Getting field dims leads to reading info from the group. If this is incomplete this may raise, with similar or different problems than that of loading the entire group.

If loading the group fails, the latest scippnexus is falling back to loading as a `DataGroup`. Similarly, this change avoids problems when trying to load fields of a broken/incomplete group.

Also includes a minor change with fewer linebreaks in warnings to improve readability. We should probably consider logging warnings instead?

### DIFF
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -362,7 +362,10 @@ class NXobject:
         if isinstance(name, str):
             item = self._group[name]
             if is_dataset(item):
-                dims = self._get_field_dims(name) if use_field_dims else None
+                try:
+                    dims = self._get_field_dims(name) if use_field_dims else None
+                except Exception:
+                    dims = None
                 return Field(item,
                              dims=dims,
                              ancestor=self,
@@ -377,9 +380,10 @@ class NXobject:
             if type(self)._getitem == NXobject._getitem:
                 raise
             else:
-                warnings.warn(
-                    f"Failed to load {self.name} as {type(self).__name__}:\n{e}\n"
+                msg = (
+                    f"Failed to load {self.name} as {type(self).__name__}: {e}. "
                     "Falling back to loading HDF5 group children as scipp.DataGroup.")
+                warnings.warn(msg)
             da = NXobject._getitem(self, name)
         if (t := self.depends_on) is not None:
 
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -364,7 +364,10 @@ class NXobject:
             if is_dataset(item):
                 try:
                     dims = self._get_field_dims(name) if use_field_dims else None
-                except Exception:
+                except Exception as e:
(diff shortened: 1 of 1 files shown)


src/scippnexus/nxobject.py
```diff
@@ -362,7 +362,10 @@ def _get_child(
         if isinstance(name, str):
             item = self._group[name]
             if is_dataset(item):
-                dims = self._get_field_dims(name) if use_field_dims else None
+                try:
+                    dims = self._get_field_dims(name) if use_field_dims else None
+                except Exception:
```
jl-wynen, line 367: Does this `except` need to be so general? Or can you list the relevant exceptions?

jl-wynen:
I'm starting to wonder how to debug this. If you try to load a file and some parts don't get loaded as you would expect, it can be very difficult to figure out why.
Would it help to log warning or info messages whenever the loader hits a fallback?

> Also includes a minor change with fewer linebreaks in warnings to improve readability. We should probably consider logging warnings instead?

Maybe. But raising warnings has the advantage that users can react to them. They can, e.g., wrap their loading code in a warning filter and error out if the file is broken. This is a lot more difficult with logging.

src/scippnexus/nxobject.py
SimonHeybrock: Maybe, but then we will likely need to revisit, since it is unlikely that I will correctly list all of them in the first attempt, since NXobject child classes may raise for a variety of reasons. `_get_field_dims` is quite horrible with a lot of nested calls, but so far I have failed to come up with a better solution.

SimonHeybrock:
> I'm starting to wonder how to debug this. If you try to load a file and some parts don't get loaded as you would expect, it can be very difficult to figure out why.
> Would it help to log warning or info messages whenever the loader hits a fallback?

We are warning when loading as a `DataGroup`, with the original exception test. In this case, the difference in the `except` branch is that dims would be default dims (`('dim_0', ...)`, and I am not sure a warning would add much?

jl-wynen:
>  and I am not sure a warning would add much?

Would it allow users to distinguish between a missing attr and a broken one?

SimonHeybrock:
> > and I am not sure a warning would add much?
> 
> Would it allow users to distinguish between a missing attr and a broken one?

Probably in some cases. There is also a bunch of guessing based on shape going on, and things may fail if, e.g., a `value` dataset is missing, etc..

I will add a warning.

src/scippnexus/nxobject.py
SimonHeybrock: I don't think so, because `_get_field_dims` is a "ScippNexus" feature, trying to give nice names to dimensions. When we default to `None` it will just load the dataset with labels `dim_0`, `dim_1`, etc., i.e., closer to what you'd get with `h5py` directly.

jl-wynen (approved):

PR #103, opened 2023-02-08
Improve dim guessing for multi-dim coords

SimonHeybrock:
Also try to broadcast errors if they are lower dimensional.

I am not sure about these changes: Strictly speaking, all this does is add support for badly written NeXus files. In particular, these changes make NeXus files *written* by Mantid read better. I am not happy with this, but I can't see much harm either.

### DIFF
--- a/src/scippnexus/nxdata.py
+++ b/src/scippnexus/nxdata.py
@@ -178,6 +178,11 @@ class NXdata(NXobject):
         if name in self._get_axes():
             # If there are named axes then items of same name are "dimension
             # coordinates", i.e., have a dim matching their name.
+            # However, if the item is not 1-D we need more labels. Try to use labels of
+            # signal if dimensionality matches.
+            if self._signal_name in self and self._get_child(name).ndim == len(
+                    self.shape):
+                return self[self._signal_name].dims
             return [name]
         return self._try_guess_dims(name)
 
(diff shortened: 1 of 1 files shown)


src/scippnexus/nxdata.py
```diff
@@ -210,7 +215,11 @@ def _getitem(self, select: ScippIndex) -> sc.DataArray:
         signal = signal[select]
         if self._errors_name is not None:
             stddevs = self[self._errors_name][select]
-            signal.variances = sc.pow(stddevs, sc.scalar(2)).values
+            # According to the standard, errors must have the same shape as the data.
+            # This is not the case in all files we observed, is there any harm in
+            # attempting a broadcast?
```
jl-wynen, line 220: Depending on what the errors are, this can hide correlations.

jl-wynen (commented):
> In particular, these changes make NeXus files written by Mantid read better. 

This seems relevant because we can only load MCStas files by first going through Mantid. At least last time I checked...

src/scippnexus/nxdata.py
SimonHeybrock: I think this is not relevant here: Mantid has an optimization, sharing the 1-D errors array (and data) for all pixels, until they are modified, using a copy-on-write mechanism. I believe this is then simply mirrored on the written NeXus file, i.e., and artifact of an implementation detail.

src/scippnexus/nxdata.py
```diff
@@ -178,6 +178,11 @@ def _get_field_dims(self, name: str) -> Union[None, List[str]]:
         if name in self._get_axes():
             # If there are named axes then items of same name are "dimension
             # coordinates", i.e., have a dim matching their name.
+            # However, if the item is not 1-D we need more labels. Try to use labels of
+            # signal if dimensionality matches.
+            if self._signal_name in self and self._get_child(name).ndim == len(
```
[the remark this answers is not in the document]
SimonHeybrock: You mean instead of `len(self.shape)`?

jl-wynen, line 183: No, like this:
```suggestion
            if self._signal_name in self and self._get_child(self._signal_name).ndim == len(
```

src/scippnexus/nxdata.py
jl-wynen: ok

src/scippnexus/nxdata.py
SimonHeybrock: No, I need to check if the requested child (`name`) has the same dims as the signal.

jl-wynen (approved):

PR #106, opened 2023-02-13
Load all items in NXtransforms groups

### LINKED ISSUE #100: Improve loading of NXtransformations
Currently, when there is a `depend_on` in, e.g., an `NXdetector`, the corresponding transformation is loaded as a Scipp `affine_transform3`. This is fine. However, if there is no `depends_on`, `NXtransformations` are just loaded as their datasets. The problem is that vital information for transformations is stored in their attributes. This is cumbersome (hard to use) and currently not supported by `scipp.Variable` or `scipp.DataGroup`.

Instead, we should load the transformations in `NXtransformations` as `scipp.Variable` with the correct spatial dtype (could be, e.g., `rotation3` or `translation3`). The code for this all exists since it is used for computing the `depends_on` transformation chain, but it is not called when loading a plain `NXtransformations` group.

jl-wynen:
Fixes #100

### DIFF
--- a/src/scippnexus/nxtransformations.py
+++ b/src/scippnexus/nxtransformations.py
@@ -1,11 +1,11 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 # @author Simon Heybrock
-from typing import Union
+from typing import List, Optional, Union
 
 import numpy as np
 import scipp as sc
-import scipp.spatial
+from prompt_toolkit.layout.processors import Transformation
 
 try:
     from scipp.scipy import interpolate
@@ -19,12 +19,23 @@ class TransformationError(Exception):
     pass
 
 
-def make_transformation(obj, /, path):
+def make_transformation(obj, /, path) -> Optional[Transformation]:
     if path.startswith('/'):
         return Transformation(obj.file[path])
     elif path != '.':
         return Transformation(obj.parent[path])
     return None  # end of chain
+
+
+class NXtransformations(NXobject):
+    """Group of transformations."""
+
+    def _getitem(self, index: ScippIndex) -> sc.DataGroup:
+        return sc.DataGroup({
+            name: get_full_transformation_starting_at(Transformation(child),
+                                                      index=index)
+            for name, child in self.items()
+        })
 
 
 class Transformation:
@@ -110,7 +121,8 @@ def _smaller_unit(a, b):
         return b.unit
 
 
-def get_full_transformation(depends_on: Field) -> Union[None, sc.DataArray]:
+def get_full_transformation(
+        depends_on: Field) -> Union[None, sc.DataArray, sc.Variable]:
     """
     Get the 4x4 transformation matrix for a component, resulting
     from the full chain of transformations linked by "depends_on"
@@ -118,7 +130,14 @@ def get_full_transformation(depends_on: Field) -> Union[None, sc.DataArray]:
     """
     if (t0 := make_transformation(depends_on, depends_on[()])) is None:
         return None
-    transformations = _get_transformations(t0)
+    return get_full_transformation_starting_at(t0)
+
+
+def get_full_transformation_starting_at(
+        t0: Transformation,
+        *,
+        index: ScippIndex = None) -> Union[None, sc.DataArray, sc.Variable]:
+    transformations = _get_transformations(t0, index=() if index is None else index)
 
     total_transform = sc.spatial.affine_transform(value=np.identity(4), unit=sc.units.m)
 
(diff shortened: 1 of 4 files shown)


tests/nxtransformations_test.py
```diff
@@ -211,3 +211,106 @@ def test_broken_time_dependent_transformation_returns_path_and_transformations(n
...
+
+
+def test_nxtransformations_group_single_item(nxroot):
+    value = sc.scalar(2.4, unit='mm')
+    offset = sc.spatial.translation(value=[6, 2, 6], unit='mm')
+    vector = sc.vector(value=[0, 1, 1])
+    t = value.to(unit='m') * vector
+    expected = (sc.spatial.translations(dims=t.dims, values=t.values, unit=t.unit) *
+                sc.spatial.translation(value=[0.006, 0.002, 0.006], unit='m') *
+                sc.spatial.affine_transform(value=np.identity(4), unit='m'))
```
SimonHeybrock, line 232: Do we want to stick closer to the dtype of the transform, instead of mapping to `affine_transform3`?

tests/nxtransformations_test.py
```diff
@@ -211,3 +211,106 @@ def test_broken_time_dependent_transformation_returns_path_and_transformations(n
...
+                 sc.spatial.translation(value=[0.006, 0.002, 0.006], unit='m') *
+                 sc.spatial.affine_transform(value=np.identity(4), unit='m'))
+
+    loaded = nxroot['transformations'][()]
+    assert set(loaded.keys()) == {'t1', 't2'}
+    assert sc.identical(loaded['t1'], expected1)
+    assert sc.allclose(loaded['t2'], expected2)
+
+
+def test_chain_with_multiple_values_asd(nxroot):
```
SimonHeybrock, line 294: What is `asd`?

tests/nxtransformations_test.py
jl-wynen: This would be inconsistent with how they are loaded, e.g., via a `depends_on` attr in a detector.
Though we can do things differently in this case. But it would make the implementation more complicated.

SimonHeybrock: > This would be inconsistent with how they are loaded, e.g., via a depends_on attr in a detector.

I wonder if this is just legacy (from the `load_nexus` implementation). Can you see any reason why we should not return the actual type (translation, rotation, or combination)?

src/scippnexus/nxtransformations.py
```diff
@@ -28,11 +28,16 @@ class NXtransformations(NXobject):
...
-            name: get_full_transformation_starting_at(Transformation(child),
-                                                      index=index)
-            for name, child in self.items()
-        })
+
+        def get_transformation(field: Field):
+            transformation = Transformation(field)
+            if 'depends_on' in field.attrs:
+                return get_full_transformation_starting_at(transformation, index=index)
+            return transformation[index]
```
SimonHeybrock, line 36: Why do we need to branch?

jl-wynen: Because `get_full_transformation_starting_at` always returns an affine transform

jl-wynen: I think changing it to this should make it so it always returns an appropriate transform:
```python
    total_transform = None

    for transform in transformations:
        if total_transform is None:
            total_transform = transform
        elif isinstance(total_transform, sc.DataArray) and isinstance(
                transform, sc.DataArray):
...
```

SimonHeybrock: I was trying to suggest above that we should consider changing that.

tests/nxtransformations_test.py
SimonHeybrock: Why do chains have to be affine?

src/scippnexus/nxtransformations.py
jl-wynen: done

SimonHeybrock: I suspect this might break (an implementation detail) of `scippneutron.load_nexus`. Can you try if it still works?

tests/nxtransformations_test.py
jl-wynen: They don't it was just implemented like this for convenience, I guess. See below

src/scippnexus/nxtransformations.py
jl-wynen: There are indeed a few test failures. Mostly from unit changes. The old code has a default unit of `m`. Is this required by load_nexus?

SimonHeybrock: I think it should be fine, provided that our coord transforms are flexible enough to deal with that? Or we convert to `m` in `load_nexus`?

jl-wynen: It looks like the tests create files with unit `dimensionless` and expect that that is loaded as `m`. There is a comment:
> Resulting position will always be in metres, whatever units are used in the NeXus file

This looks like something `load_nexus` should enforce and not ScippNeXus. Unless the standard says that the unit defaults to `m`. Do you know?

SimonHeybrock: I can't remember. I agree that ScippNexus should not do this.

SimonHeybrock (approved):

PR #109, opened 2023-02-16
Improve shape loading

SimonHeybrock:
Improves loading of `NXoff_geometry` and `NXcylindrical_geometry`.

When such a group is loaded directly, the fields are loaded as-is. When embedded in, e.g., an NXdetector, its detector_numbers may be used to derive a shape for each detector pixel. This ensuresthat this can be tracked and broadcasted (e.g., against pixel offsets) properly, avoiding need to wrap in a scalar variable, which would prevent this.

This work has highlighted the need for butter support for "structured" dtypes in Scipp, see scipp/scipp#3046. For now, this is using binned data for a similar (if slightly more restricted) approach.

### DIFF
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -399,20 +399,14 @@ class NXobject:
                 container[name] = obj
 
         try:
+            from .nxcylindrical_geometry import NXcylindrical_geometry
+            from .nxoff_geometry import NXoff_geometry
             da = self._getitem(name)
-            from .nxoff_geometry import NXoff_geometry, off_to_shape
-
-            # TODO if loads as DataGroup (NXsample?), do we need to wrap in binned?
-            # cases:
-            # DataArray/DataGroup
-            # detector_numbers or not (never detector numbers in DataGroup case?)
-            # scalar shape
-            for key, off in self[NXoff_geometry].items():
-                if (detnum := getattr(self, 'detector_number', None)) is not None:
-                    detnum = da.coords[detnum]
-                    insert(da, key, off_to_shape(**off[()], detector_number=detnum))
-                else:
-                    da[key] = off_to_shape(**off[()])
+            for key, child in self[[NXcylindrical_geometry, NXoff_geometry]].items():
+                detector_number = getattr(self, 'detector_number', None)
+                if detector_number is not None:
+                    detector_number = da.coords[detector_number]
+                insert(da, key, child.load_as_array(detector_number=detector_number))
         except Exception as e:
             # If the child class cannot load this group, we fall back to returning the
             # underlying datasets in a DataGroup.
--- a/src/scippnexus/nxobject.py
+++ b/src/scippnexus/nxobject.py
@@ -419,9 +419,13 @@ class NXobject:
             # loaded.
             if isinstance(t, str):
                 from .nexus_classes import NXtransformations
-                for name, group in self[NXtransformations].items():
-                    insert(name, group[()])
-
+                for key, group in self[NXtransformations].items():
+                    insert(key, group[()])
+        from .nxoff_geometry import NXoff_geometry, off_to_shape
+        for key, off in self[NXoff_geometry].items():
+            detector_number = getattr(self, 'detector_number', None)
+            detector_number = da.coords[detector_number]
+            da.coords[key] = off_to_shape(**off[()], detector_number=detector_number)
         return da
 
     def _get_children_by_nx_class(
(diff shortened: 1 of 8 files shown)


tests/nxdetector_test.py
```diff
@@ -304,3 +313,166 @@ def __getitem__(self, select=...):
...
+    da.coords['xx'] = sc.array(dims=['xx'], unit='m', values=[0.1, 0.2])
+    detector = nxroot.create_class('detector0', NXdetector)
+    detector.create_field(detid_name, da.coords['detector_number'])
+    detector.create_field('xx', da.coords['xx'])
+    detector.create_field('data', da.data)
+    detector.attrs['axes'] = ['xx', 'yy']
+    create_off_geometry_detector_numbers_1234(detector, name='shape')
+    loaded = detector[...]
+    assert sc.identical(loaded.coords['shape'].bins.size(),
+                        sc.array(dims=da.dims, values=[[1, 1], [1, 1]], unit=None))
```
jl-wynen, line 351: Can you test that the loaded data are correct instead of just that there is the expected amount of data?

tests/nxdetector_test.py
```diff
@@ -304,3 +313,166 @@ def __getitem__(self, select=...):
...
+
+
+def test_missing_detector_numbers_triggers_fallback_given_off_geometry_with_det_faces(
+        nxroot):
+    var = sc.array(dims=['xx', 'yy'], unit='K', values=[[1.1, 2.2], [3.3, 4.4]])
+    detector = nxroot.create_class('detector0', NXdetector)
+    detector.create_field('data', var)
+    detector.attrs['axes'] = ['xx', 'yy']
+    create_off_geometry_detector_numbers_1234(detector, name='shape')
+    assert isinstance(detector[...], sc.DataGroup)
```
jl-wynen, line 361: Please also test that the loaded data group contains the geometry fields.

src/scippnexus/nxobject.py
```diff
@@ -390,8 +390,23 @@ def _get_child(
...
+
         try:
+            from .nxcylindrical_geometry import NXcylindrical_geometry
+            from .nxoff_geometry import NXoff_geometry
             da = self._getitem(name)
+            for key, child in self[[NXcylindrical_geometry, NXoff_geometry]].items():
+                detector_number = getattr(self, 'detector_number', None)
+                if detector_number is not None:
+                    detector_number = da.coords[detector_number]
+                insert(da, key, child.load_as_array(detector_number=detector_number))
```
jl-wynen, line 409: You can pull 
```python
                detector_number = getattr(self, 'detector_number', None)
                if detector_number is not None:
                    detector_number = da.coords[detector_number]
```
out of the loop.

Unless `da.coords[detector_number]` can fail. In which case this failure would currently not happen if there are no children to iterate over. Can it fail?

Also, this code really wants to be its own function

src/scippnexus/nxoff_geometry.py
```diff
@@ -1,15 +1,49 @@
...
+                 vertices: sc.Variable,
+                 winding_order: sc.Variable,
+                 faces: sc.Variable,
+                 detector_faces: Optional[sc.Variable] = None,
+                 detector_number: Optional[sc.Variable] = None) -> sc.Variable:
+    """
+    Convert OFF shape description to simpler shape representation.
+    """
+    vw = vertices[winding_order.values]
+    fvw = sc.bins(begin=faces, data=vw, dim=vw.dim)
```
jl-wynen, line 21: What do these variable names mean?

src/scippnexus/nxoff_geometry.py
```diff
@@ -1,15 +1,49 @@
...
+    low = fvw.bins.size().min().value
+    high = fvw.bins.size().max().value
+    if low == high:
+        fvw = vw.fold(dim=vertices.dim, sizes={faces.dim: -1, vertices.dim: low})
+    else:
+        raise NotImplementedError("Conversion from OFF to shape not implemented for "
+                                  "inconsistent number of vertices in faces.")
+    if detector_faces is None:  # if detector_number is not None, all have same shape
+        return sc.bins(begin=sc.index(0), dim=faces.dim, data=fvw)
```
jl-wynen, line 30: What does this line do? It seems to bin already binned data.

SimonHeybrock: It wasn't, just a bad reuse of variable name, will change.

SimonHeybrock:
I think I addressed all of your comments, please have another look.

jl-wynen (approved):

PR #111, opened 2023-02-20
Handle deprecated NXgeometry like new geometry classes

SimonHeybrock:


src/scippnexus/nxobject.py
```diff
@@ -423,6 +424,8 @@ def insert(container, name, obj):
             detector_number = container.coords[detector_number]
         for key, child in self[[NXcylindrical_geometry, NXoff_geometry]].items():
             insert(container, key, child.load_as_array(detector_number=detector_number))
+        for key, child in self[NXgeometry].items():
+            insert(container, key, child[()])
```
jl-wynen, line 428: Why does this use `child[()]` and not `child.load_as_array` like the other geometry classes?

jl-wynen (commented):
Got any files where this is used?

src/scippnexus/nxobject.py
SimonHeybrock: Because we are not performing any transformations on it, it is just loaded as a plain `DataGroup`.

SimonHeybrock:
> Got any files where this is used?

There is a test in scippneutron that loads a PG3 file that uses this.

jl-wynen (approved):

PR #112, opened 2023-02-21
pyproject.toml missing in conda test env

SimonHeybrock:
This meant that pytest options were not used in the 'test' step of conda build.


jl-wynen (approved):
This also affects our other projects, right?

SimonHeybrock:
Yes, looks like it. I think it never mattered because pytest ran fine without config (but it would, e.g., not get the desired "warnings as errors" config param).

PR #113, opened 2023-02-21
More refined exception handling for fallback loader

### LINKED ISSUE #107: More selective fallback loader
Currently, the fallback loader in `NXobject` catches (nearly) all exceptions from loaders for concrete classes and uses the fallback. This is intended to allow loading files with partially bad structure. But it also hides user errors like a bad index (wrong dim, bad slice, etc.).

We should distinguish between errors originating in the file structure and error originating from the user/caller. Only the former should trigger the fallback.

SimonHeybrock:
I am pretty sure this is not the last we will see of this: There are probably more places in the NXobject child classes that need to raise the right exception type. For now this works on the files in `pytest -m externalfiles` and a couple of others.

Fixes #107

### DIFF
--- a/src/scippnexus/_common.py
+++ b/src/scippnexus/_common.py
@@ -56,7 +56,7 @@ def _to_canonical_select(dims: List[str],
 
     def check_1d():
         if len(dims) != 1:
-            raise ValueError(f"Dataset has multiple dimensions {dims}, "
+            raise IndexError(f"Dataset has multiple dimensions {dims}, "
                              "specify the dimension to index.")
 
     if select is Ellipsis:
@@ -69,14 +69,14 @@ def _to_canonical_select(dims: List[str],
     if isinstance(select, tuple):
         check_1d()
         if len(select) != 1:
-            raise ValueError(f"Dataset has single dimension {dims}, "
+            raise IndexError(f"Dataset has single dimension {dims}, "
                              "but multiple indices {select} were specified.")
         return {dims[0]: select[0]}
     elif isinstance(select, int) or isinstance(select, slice):
         check_1d()
         return {dims[0]: select}
     if not isinstance(select, dict):
-        raise ValueError(f"Cannot process index {select}.")
+        raise IndexError(f"Cannot process index {select}.")
     return select
 
 
@@ -88,7 +88,7 @@ def to_plain_index(dims: List[str], select: ScippIndex) -> Union[int, slice, tup
     index = [slice(None)] * len(dims)
     for key, sel in select.items():
         if key not in dims:
-            raise ValueError(
+            raise IndexError(
                 f"'{key}' used for indexing not found in dataset dims {dims}.")
         index[dims.index(key)] = sel
     if len(index) == 1:
--- a/src/scippnexus/_common.py
+++ b/src/scippnexus/_common.py
@@ -56,8 +56,8 @@ def _to_canonical_select(dims: List[str],
 
     def check_1d():
         if len(dims) != 1:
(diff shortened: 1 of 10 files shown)


src/scippnexus/_common.py
```diff
@@ -56,7 +56,7 @@ def _to_canonical_select(dims: List[str],
 
     def check_1d():
         if len(dims) != 1:
-            raise ValueError(f"Dataset has multiple dimensions {dims}, "
+            raise IndexError(f"Dataset has multiple dimensions {dims}, "
```
jl-wynen, line 59: Is this consistent with scipp or should it raise `sc.DimensionError`?

src/scippnexus/nxdata.py
```diff
@@ -137,6 +137,9 @@ def _signal(self) -> Union[Field, '_EventField', None]:  # noqa: F821
         if self._signal_override is not None:
             return self._signal_override
         if self._signal_name is not None:
+            if self._signal_name not in self:
+                raise NexusStructureError(
+                    f"Signal field '{self._signal_name}' not found in group.")
```
jl-wynen, line 142: Could use `try-except` to translate the exception. It's kind of a question of style. But you did this elsewhere.

src/scippnexus/nxtransformations.py
```diff
@@ -28,11 +28,15 @@ class NXtransformations(NXobject):
...
-                                                      index=index)
-            for name, child in self.items()
-        })
+        try:
+            return sc.DataGroup({
+                name: get_full_transformation_starting_at(Transformation(child),
+                                                          index=index)
+                for name, child in self.items()
+            })
+        except (sc.DimensionError, sc.UnitError) as e:
```
jl-wynen, line 37: Can a `UnitError` be raised when a user uses label-based indexing with a bad unit?

src/scippnexus/_common.py
SimonHeybrock: I felt that always raising `IndexError` makes most sense here, since it tells the user "you passes a bad index". But I may be convinced otherwise?

Consistency with scipp would make sense, but maybe it is scipp that should change? Or should we raise `IndexError` only for "out of range" and `DimensionError` otherwise?

src/scippnexus/nxtransformations.py
SimonHeybrock: ScippNexus is not support label-based indexing currently.

src/scippnexus/_common.py
jl-wynen: https://docs.python.org/3/library/exceptions.html?highlight=indexerror#IndexError
> Raised when a sequence subscript is out of range. (Slice indices are silently truncated to fall in the allowed range; if an index is not an integer, [TypeError](https://docs.python.org/3/library/exceptions.html?highlight=indexerror#TypeError) is raised.)

So I would say we use `DimensionError`.

src/scippnexus/nxtransformations.py
jl-wynen: So it might need to change in the future if this is added. But It seems difficult to fix. So leave it for now.

SimonHeybrock: Hmm, now that we raise `DimensionError` instead of `IndexError` I cannot distinguish any more between errors from a user selection and errors from dim mismatches in the transforms in the file.

jl-wynen: So the transformation code would need an explicit handling and raise `NeXusStructureError`?

jl-wynen (approved):

PR #123, opened 2023-04-05
Improve handling of event data in NXdetector

SimonHeybrock:
This resolves a number of inconsistencies and broken things. For example, slicing pixels works now. Behavior wise, we are now closer to the pre-v2 implementation.

When an NXdetector contains both dense data and events, this is not returning a `Dataset`. I don't know if this is useful in practice, or if we should just go back to returning the events by default?

### DIFF
--- a/src/scippnexus/v2/nxdata.py
+++ b/src/scippnexus/v2/nxdata.py
@@ -247,6 +247,8 @@ class NXdata(NXobject):
         aux = {name: dg.pop(name) for name in self._aux_signals}
         signal = dg.pop(self._signal_name)
         coords = dg
+        if isinstance(signal, sc.DataGroup):
+            raise NexusStructureError("Signal is not an array-like.")
         da = sc.DataArray(data=signal) if isinstance(signal, sc.Variable) else signal
         da = self._add_coords(da, coords)
         if aux:
@@ -325,9 +327,10 @@ def _find_embedded_nxevent_data(
 
 class EventField:
 
-    def __init__(self, event_data: Group, detector_number: Field) -> None:
+    def __init__(self, event_data: Group, grouping_name: str, grouping: Field) -> None:
         self._event_data = event_data
-        self._detector_number = detector_number
+        self._grouping_name = grouping_name
+        self._grouping = grouping
 
     @property
     def attrs(self) -> Dict[str, Any]:
@@ -335,20 +338,23 @@ class EventField:
 
     @property
     def sizes(self) -> Dict[str, int]:
-        sizes = self._event_data.sizes
-        sizes.update(self._detector_number.sizes)
+        sizes = dict(self._grouping.sizes)
+        sizes.update(self._event_data.sizes)
         return sizes
 
     @property
     def dims(self) -> Tuple[str, ...]:
-        return self._detector_number.dims + self._event_data.dims
+        return self._grouping.dims + self._event_data.dims
 
     def __getitem__(self, sel: ScippIndex) -> sc.DataArray:
         event_sel = to_child_select(self.dims, self._event_data.dims, sel)
         events = self._event_data[event_sel]
-        detector_sel = to_child_select(self.dims, self._detector_number.dims, sel)
-        return _group_events(event_data=events,
-                             grouping=self._detector_number[detector_sel])
+        detector_sel = to_child_select(self.dims, self._grouping.dims, sel)
+        if not isinstance(events, sc.DataArray):
+            return events
+        da = _group_events(event_data=events, grouping=self._grouping[detector_sel])
+        da.coords[self._grouping_name] = da.coords.pop('event_id')
+        return da
 
 
 class NXdetector(NXdata):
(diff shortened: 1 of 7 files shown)


jl-wynen:
> I don't know if this is useful in practice, or if we should just go back to returning the events by default?

Is the dense data typically used as a histogram of the events? If so, there should be an option to select which one to load. Or at least a postprocessing function.

src/scippnexus/v2/nxdata.py
```diff
@@ -84,15 +86,24 @@ def _init_signal(self, name: Optional[str], children):
         if name is not None and name in children:
             self._signal_name = name
             self._signal = children[name]
-        else:
```
jl-wynen, line 87: This seems fishy. There are now 3 blocks that each can set `self._signal` regardless of whether it is already set. So if more than one signal-detection method succeeds, only the result of the last will be used.
Is this intentional?

src/scippnexus/v2/nxdata.py
```diff
@@ -310,6 +331,38 @@ def _find_embedded_nxevent_data(
...
+
+    @property
+    def attrs(self) -> Dict[str, Any]:
+        return self._event_data.attrs
+
+    @property
+    def sizes(self) -> Dict[str, int]:
+        sizes = dict(self._grouping.sizes)
+        sizes.update(self._event_data.sizes)
+        return sizes
```
jl-wynen, line 349:
```suggestion
        return {**self._grouping.sizes, **self._event_data.sizes}
```

src/scippnexus/v2/nxdata.py
```diff
@@ -310,6 +331,38 @@ def _find_embedded_nxevent_data(
         return event_group
 
 
+class EventField:
+
+    def __init__(self, event_data: Group, grouping_name: str, grouping: Field) -> None:
+        self._event_data = event_data
+        self._grouping_name = grouping_name
+        self._grouping = grouping
```
jl-wynen, line 339: Can you explain what 'grouping' means?

src/scippnexus/v2/nxdata.py
SimonHeybrock: Yes this is intentional and I fear also required (even though I am far from happy about this), and not new (just added one more case, in a sense). For example, McStas NeXus files use the "legacy" mechanism for defining the signal. However, for some reason (bad luck?) they also define the non-legacy `signal` group-attr, but with an unrelated value. Therefore, the search continues, even if that is found.

src/scippnexus/v2/nxdata.py
SimonHeybrock: Added a docstring.

SimonHeybrock:
> 



> > I don't know if this is useful in practice, or if we should just go back to returning the events by default?
> 
> Is the dense data typically used as a histogram of the events? If so, there should be an option to select which one to load. Or at least a postprocessing function.

I *guess* the the dense data is typically just meant as a "preview" of what you would get when histogramming the events. But good luck finding that documented anywhere. It is not even clear if having both is legal.

src/scippnexus/v2/nxdata.py
jl-wynen: So can all 3 methods appear in the same file? Or only the 2 that McStas uses?

Can you reverse the order and return after the first signal was found? This would make it clearer that there are multiple, independent encodings. As it stands, one has to read everything in detail to understand that they don't depend on each other.

jl-wynen: And can you log which one gets used to help with debugging?

jl-wynen:
> > > I don't know if this is useful in practice, or if we should just go back to returning the events by default?
> > 
> > 
> > Is the dense data typically used as a histogram of the events? If so, there should be an option to select which one to load. Or at least a postprocessing function.
> 
> I _guess_ that the dense data is typically just meant as a "preview" of what you would get when histogramming the events. But good luck finding that documented anywhere. It is not even clear if having both is legal.

Based on your answer, I suspect that it will be most useful to return the events as a data array and log that there was dense data that was not read. But this is just a hunch

src/scippnexus/v2/nxdata.py
SimonHeybrock: > So can all 3 methods appear in the same file? Or only the 2 that McStas uses?

In principle anything can happen. You could take completely unrelated files and merge them into one, having multiple NXentry groups. As far as I know there is no restriction about which versions or features of the NeXus standard is used.

jl-wynen: > ScippNexus does not log. Printing warnings is also quite user-unfriendly, as there can easily be hundreds of them.

But that could change. And I think it should given that it makes a number of decisions about where to load data from.
If you are worried about noise, I think logging at debug level would be enough here.

SimonHeybrock: > Can you reverse the order and return after the first signal was found?

One could reverse the first 2 (the third finds also aux_signals), but not return (since we always run the third). But then I need to add an extra check in the non-legacy one, to check if a legacy signal was found. Doesn't seem more readable.

jl-wynen: You can run the 3rd first. If it found a signal, return, otherwise, continue. This would populate the aux signals.

SimonHeybrock: I can move the search for NXlog and NXevent_data to the top, and change the behavior to not load dense data when both events and dense is found, as discussed elsewhere.

SimonHeybrock: But I don't know if that even makes sense: I don't think the NeXus standard says anywhere that any NXlog or NXevent_data should be considered as signal, without having the `signal` attribute set in the group or subgroup. That is just our interpretation, probably?

jl-wynen: > Are you suggesting to change the current behavior?

No. Why would this change behaviour? From what I understood, the 3rd block will override the signal set earlier (if it finds one). And it will always set aux signals. So moving this to the top and returning if it find a signal should be the same behaviour, no?

jl-wynen: Now I see. How about
```python
self._aux_signals = self._find_aux_signals()
if self._signal is None:
  self._signal_name = self._aux_signals.pop(0)  # or w/o 0 if the order is irrelevant
  self._signal = children[self._signal_name]
```
to make it more explicit?

I keep asking for changes because I am, as you can tell, confused by this function and the precise logic behind it.

SimonHeybrock: I am considering changing the mechanism such that a user has to select via definitions which "version" of the standard they want to use. The would prevent support for mixed files though.

SimonHeybrock: I will add comments on #126, as that changed this function some more.

PR #126, opened 2023-04-11
Fix exception when loading NXdata with NXlog signal

SimonHeybrock:
Also affected NXdata subclasses


jl-wynen (approved):
Ok. But see the 2nd part of https://github.com/scipp/scippnexus/pull/123#discussion_r1168364691

PR #133, opened 2023-04-19
Fix issue with dims setup for NXdetector with event data

SimonHeybrock:
NXdetector initialized the dims of the detector_number field, but NXdata subsequently called NXobject.__init__, which overrode this.

Furthermore, the dim guessing logic ignored event-signals.

After this, the AMOR file now correctly loads the NXdetector as a DataArray.

### DIFF
--- a/tests/load_files_test.py
+++ b/tests/load_files_test.py
@@ -17,6 +17,7 @@ all_files = [
     '2023/NMX_2e11-rechunk.h5',
     '2023/YMIR_038243_00010244.hdf',
     '2023/amor2020n000346_tweaked.nxs',
+    '2023/LOKI_60322-2022-03-02_2205_fixed.nxs',
 ]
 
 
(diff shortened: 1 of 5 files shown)


tests/load_files_test.py
```diff
@@ -14,9 +16,48 @@
     '2023/LOKI_mcstas_nexus_geometry.nxs',
     '2023/NMX_2e11-rechunk.h5',
     '2023/YMIR_038243_00010244.hdf',
+    '2023/amor2020n000346_tweaked.nxs',
+    '2023/LOKI_60322-2022-03-02_2205_fixed.nxs',
 ]
 
 
+def get_item_at_path(dg: sc.DataGroup, path: str) -> sc.DataArray:
```
jl-wynen, line 24: Looks useful. Should this be turned into a method on `NXdata`?

tests/load_files_test.py
```diff
@@ -14,9 +16,48 @@
...
+    nodes = path.split('/')
+    for node in nodes:
+        dg = dg[node]
+    return dg
+
+
+def assert_schema(dg: sc.DataGroup, schema: Dict[str, Any]) -> None:
```
jl-wynen, line 31: Do you want to use this for the 2 old tests as well?

jl-wynen (approved):

tests/load_files_test.py
SimonHeybrock: Note sure what you mean, `scippnexus.Group` supports path-based access just like `h5py`.

tests/load_files_test.py
SimonHeybrock: For now I wanted to ensure that we can load the ESS files we care about correctly. Some of the older files are quite broken (we have open issues with ECDC), so this may not be so useful there since fallback to DataGroup is triggered in many cases.

jl-wynen (approved):

PR #137, opened 2023-04-24
Fix event selection of length 0

### LINKED ISSUE #136: Loading empty `event_time_zero` slice seems to load everything
```python
import scippnexus.v2 as snx
# ...
dg = f['event_time_zero', 0:1]  # ok, loads one pulse
dg = f['event_time_zero', 0:0]  # seems to load everything?
```

SimonHeybrock:
Previously this setup correct empty pulse bins, but loaded the entire event buffer. When binning into detectors the pulse binning is ignored nd we got back all events.

Fixes #136.

### DIFF
--- a/tests/nxdetector_test.py
+++ b/tests/nxdetector_test.py
@@ -326,14 +326,16 @@ def test_event_data_field_dims_labels(nxroot):
     assert detector['detector_number'].dims == ('detector_number', )
 
 
-def test_nxevent_data_selection_yields_correct_pulses(nxroot):
+def test_nxevent_data_without_detector_number_selection_yields_correct_pulses(nxroot):
     detector = nxroot.create_class('detector0', NXdetector)
     create_event_data_ids_1234(detector.create_class('events', snx.NXevent_data))
 
     class Load:
 
         def __getitem__(self, select=...):
-            da = detector['events'][select]
+            da = detector[select]
+            assert da.bins.size().sum(
+            ).value == da.bins.constituents['data'].sizes['event']
             return da.bins.size().values
 
     assert np.array_equal(Load()[...], [3, 0, 2, 1])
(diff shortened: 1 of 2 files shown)


tests/nxdetector_test.py
```diff
@@ -355,6 +357,38 @@ def __getitem__(self, select=...):
...
+            return da.bins.size().values
+
+    assert np.array_equal(Load()[...], [2, 3, 0, 1])
+    assert np.array_equal(Load()['event_time_zero', 0], [1, 1, 0, 1])
+    assert np.array_equal(Load()['event_time_zero', 1], [0, 0, 0, 0])
+    assert np.array_equal(Load()['event_time_zero', 2], [1, 1, 0, 0])
+    assert np.array_equal(Load()['event_time_zero', 3], [0, 1, 0, 0])
+    assert np.array_equal(Load()['event_time_zero', -1], [0, 1, 0, 0])
+    assert np.array_equal(Load()['event_time_zero', -2], [1, 1, 0, 0])
+    assert np.array_equal(Load()['event_time_zero', 0:0], [0, 0, 0, 0])
```
nvaytet, line 379: Test looks good. Are you saying that before you made the fix, here you would get `[2, 3, 0, 1]` instead of `[0, 0, 0, 0]`?

nvaytet (approved):

tests/nxdetector_test.py
SimonHeybrock: Yes exactly, that is what happened.

SimonHeybrock: The interesting bit is that (as you can see above) there was an actual unit test for this case. The problem was that it only checked that 0 pulses where loaded from NXevent_data.... but it did not verify that the length of the content buffer was also zero. Thus, when binning into detectors the events magically reappeared.

PR #139, opened 2023-05-01
Fall back to dim labels from HDF5

SimonHeybrock:
See code comments.

### DIFF
--- a/src/scippnexus/v2/nxdata.py
+++ b/src/scippnexus/v2/nxdata.py
@@ -217,7 +217,7 @@ class NXdata(NXobject):
             for key, attr in attrs.items() if key.endswith(indices_suffix)
         }
 
-        dims = np.array(group_dims)
+        dims = np.array(self._axes)
         self._dims_from_indices = {
             key: tuple(dims[np.array(indices).flatten()])
             for key, indices in indices_attrs.items()
@@ -231,6 +231,10 @@ class NXdata(NXobject):
             return self._group_dims
         # Latest way of defining dims
         if (dims := self._dims_from_indices.get(name)) is not None:
+            if '.' in dims:
+                hdf5_dims = self._dims_from_hdf5(field)
+                return tuple(dim if dim != '.' else hdf5_dim
+                             for dim, hdf5_dim in zip(dims, hdf5_dims))
             return dims
         # Older way of defining dims via axis attribute
         if (axis := self._axis_index.get(name)) is not None:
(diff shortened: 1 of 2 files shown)


src/scippnexus/v2/nxdata.py
```diff
@@ -231,6 +231,10 @@ def _get_dims(self, name, field):
             return self._group_dims
         # Latest way of defining dims
         if (dims := self._dims_from_indices.get(name)) is not None:
+            if '.' in dims:
+                hdf5_dims = self._dims_from_hdf5(field)
+                return tuple(dim if dim != '.' else hdf5_dim
+                             for dim, hdf5_dim in zip(dims, hdf5_dims))
```
jl-wynen, line 237: Is there a risk that this skips dim labels? (Because `_dims_from_hdf5` drops trailing empty dims.)

src/scippnexus/v2/nxdata.py
```diff
@@ -252,6 +256,19 @@ def _get_dims(self, name, field):
...
+        # While not mandated or recommended by the standard, we can try to find HDF5
+        # dim labels as a fallback option for defining dimension labels. Ideally we
+        # would like to do so in NXobject._init_field, but this causes significant
+        # overhead for small files with many datasets. Defined here, this will only
+        # take effect for NXdata, NXdetector, NXlog, and NXmonitor.
+        return self._dims_from_hdf5(field)
+
+    def _dims_from_hdf5(self, field):
+        hdf5_dims = [dim.label for dim in getattr(field.dataset, 'dims', [])]
+        if any([dim != '' for dim in hdf5_dims]):
```
jl-wynen, line 268:
```suggestion
        if any(dim != '' for dim in hdf5_dims):
```

src/scippnexus/v2/nxdata.py
SimonHeybrock: You may be right, good catch.

SimonHeybrock: After double checking, this should be ok, and desired. See how this function is used:

```python
        elif (dims := self._get_dims(name, field)) is not None:
            # The convention here is that the given dimensions apply to the shapes
            # starting from the left. So we only squeeze dimensions that are after
            # len(dims).
            shape = _squeeze_trailing(dims, field.dataset.shape)
            field.sizes = dict(zip(dims, shape))
```

jl-wynen (approved):

PR #145, opened 2023-05-24
Fix two bugs

SimonHeybrock:
See commits. Making a patch release for this.

See https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-data-axes-attribute for reference on the comma->colon fix.

### DIFF
--- a/src/scippnexus/nxdata.py
+++ b/src/scippnexus/nxdata.py
@@ -158,7 +158,7 @@ class NXdata(NXobject):
                 # The standard says that the axes should be colon-separated, but some
                 # files use comma-separated.
                 if len(dims) == 1 and self._signal.ndim > 1:
-                    self._dims = tuple(axes.split(','))
+                    dims = tuple(axes.split(','))
                 return dims
         return []
 
--- a/src/scippnexus/nxdata.py
+++ b/src/scippnexus/nxdata.py
@@ -154,7 +154,12 @@ class NXdata(NXobject):
             return [a for a in axes if a != '.']
         elif (signal := self._signal) is not None:
             if (axes := signal.attrs.get('axes')) is not None:
-                return axes.split(':')
+                dims = axes.split(':')
+                # The standard says that the axes should be colon-separated, but some
+                # files use comma-separated.
+                if len(dims) == 1 and self._signal.ndim > 1:
+                    self._dims = tuple(axes.split(','))
+                return dims
         return []
 
     def _guess_dims(self, name: str):
--- a/src/scippnexus/nxdata.py
+++ b/src/scippnexus/nxdata.py
@@ -246,7 +246,11 @@ class NXdata(NXobject):
                 sc.pow(stddevs, sc.scalar(2)).values, shape=signal.shape
             )
 
-        da = sc.DataArray(data=signal) if isinstance(signal, sc.Variable) else signal
+        da = (
+            signal
+            if isinstance(signal, sc.DataArray)
+            else sc.DataArray(data=asarray(signal))
+        )
 
         skip = self._skip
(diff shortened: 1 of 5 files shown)


SimonHeybrock:
@jl-wynen I'll be merging this to make a patch release, please review nevertheless.

src/scippnexus/nxdata.py
```diff
@@ -246,7 +246,11 @@ def _getitem(self, select: ScippIndex) -> sc.DataArray:
                 sc.pow(stddevs, sc.scalar(2)).values, shape=signal.shape
             )
 
-        da = sc.DataArray(data=signal) if isinstance(signal, sc.Variable) else signal
+        da = (
+            signal
+            if isinstance(signal, sc.DataArray)
+            else sc.DataArray(data=asarray(signal))
+        )
```
jl-wynen, line 258: Would it make sense if the DataArray constructor behaved in this way automatically? I.e. `sc.DataArray(signal)` would be equivalent to this code. This is how many builtin containers behave to allow users to ensure that something is, e.g., a list without having to know the precise input type.

Not for this PR but in genera...

src/scippnexus/nxdata.py
```diff
@@ -154,7 +154,12 @@ def _get_axes(self):
             return [a for a in axes if a != '.']
         elif (signal := self._signal) is not None:
             if (axes := signal.attrs.get('axes')) is not None:
-                return axes.split(':')
+                dims = axes.split(':')
+                # The standard says that the axes should be colon-separated, but some
+                # files use comma-separated.
+                if len(dims) == 1 and self._signal.ndim > 1:
+                    self._dims = tuple(axes.split(','))
+                return dims
```
jl-wynen, line 162: Why does this set `self._dims` in the comma case but not the colon case?

src/scippnexus/v2/nxdata.py
```diff
@@ -151,6 +151,10 @@ def _init_axes(
         )
         if self._signal_axes is not None:
             self._signal_axes = tuple(self._signal_axes.split(':'))
+            # The standard says that the axes should be colon-separated, but some
+            # files use comma-separated.
+            if len(self._signal_axes) == 1 and self._signal.dataset.ndim > 1:
+                self._signal_axes = tuple(self._signal_axes[0].split(','))
```
jl-wynen, line 157: Add a helper function since this code appears 3 times and is apparently non trivial?

src/scippnexus/nxdata.py
SimonHeybrock: Maybe?

src/scippnexus/v2/nxdata.py
SimonHeybrock: No, because 2 of those will be removed when we make `v2` the default

src/scippnexus/nxdata.py
```diff
@@ -154,7 +154,12 @@ def _get_axes(self):
             return [a for a in axes if a != '.']
         elif (signal := self._signal) is not None:
             if (axes := signal.attrs.get('axes')) is not None:
-                return axes.split(':')
+                dims = axes.split(':')
+                # The standard says that the axes should be colon-separated, but some
+                # files use comma-separated.
+                if len(dims) == 1 and self._signal.ndim > 1:
+                    self._dims = tuple(axes.split(','))
```
SimonHeybrock, line 161:
```suggestion
                    dims = tuple(axes.split(','))
```

src/scippnexus/nxdata.py
SimonHeybrock: Fixed.

jl-wynen (approved):

PR #154, opened 2023-07-07
Use alignment flag in slicing instead of attrs

### LINKED ISSUE #150: Switch to using alignment flags instead of attrs

jl-wynen:
Fixes #150

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,19 +10,19 @@ confuse==2.0.0
 graphlib-backport==1.0.3
     # via scipp
 h5py==3.8.0
-    # via -r base.in
+    # via -r requirements/base.in
 numpy==1.24.1
     # via
     #   h5py
     #   scipp
     #   scipy
 python-dateutil==2.8.2
-    # via -r base.in
+    # via -r requirements/base.in
 pyyaml==6.0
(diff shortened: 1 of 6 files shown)


requirements/base.txt
```diff
@@ -5,24 +5,24 @@
...
 pyyaml==6.0
     # via confuse
-scipp==23.5.0
-    # via -r base.in
-scipy==1.10.0
-    # via -r base.in
+scipp==23.7.0
+    # via -r requirements/base.in
+scipy==1.11.1
+    # via -r requirements/base.in
```
SimonHeybrock, line 26: Generally, please avoid running global dependency updates alongside other changes.

SimonHeybrock (commented):
Is there no test that needs updating for the new behavior?

requirements/base.txt
jl-wynen: Replaced it with updating scipp only

jl-wynen:
> Is there no test that needs updating for the new behavior?

Don't think so. There are tests that fail with the latest Scipp version and without this update. They compare slicing in ScippNeXus directly with slicing in Scipp.

SimonHeybrock (approved):

PR #164, opened 2023-09-20
fix: incorrect annotation

### LINKED ISSUE #151: Incorrect type in protocols
`H5Base.attrs` is annotated to return `List[int]` but the actual implementations return `Attrs` (dict-like).

jokasimr:
Fixes #151.


YooSunYoung:
We just tried static type-checker, and I was not sure if `H5Base.attr` should be annotated as `Mapping` or `Attr`.

YooSunYoung (approved):

PR #166, opened 2023-09-26
Read `pixel_masks` from NXdetector

### LINKED ISSUE #157: Load NXdetector pixel_mask into DataArray.masks
See https://manual.nexusformat.org/classes/base_classes/NXdetector.html#nxdetector-pixel-mask-field. We should turn this into masks of a `scipp.DataArray` (the the detector is loaded as a DataArray).

Scipp only really works with boolean masks. Therefore, we should inspect the bitmask, and split it into individual masks. Only bits that are actually in use should result in creation of a corresponding mask.

jokasimr:
Fixes #157.

~This is still a very rough draft.~
~I need a nexus file with a pixel mask to experiment with.~

The transformation from bitmask to dictionary of masks is tested.
But actually reading masks from a nexus file is still untested.


src/scippnexus/nxdata.py
```diff
@@ -579,6 +579,28 @@ def _maybe_event_field(name: str, child: Union[Field, Group]):
...
+            2: 'under_responding',
+            3: 'virtual_pixel',
+            4: 'noisy',
+            6: 'pixel_is_part_of_a_cluster_of_problematic_pixels',
+            8: 'user_defined_mask',
+            31: 'virtual_pixel',
+        }
+        masks = {
+            name: (nxmask >> (31 - bit)) % 2 for bit, name in bit_to_mask_name.items()
+        }
```
SimonHeybrock, line 601: Should add a condition so only masks that are actually used are added?

SimonHeybrock (commented):
I think we need to test this without files anyway, so I suggest for the time being we just write a unit test with fake data (like most of our other tests). If actual files are different from our understanding we can fix it later.

SimonHeybrock:
We should also decide how to behave when one of the "undefined" bits is set? Should it raise? Save to mask an an integer dtype-mask (which won't work in operations)? Use the bit offset as mask name?

SimonHeybrock:
> How do we determine which bits are in use? If it is set to 0 in the entire array?

There is a number of ways to do this. One could use `min` and `max` and test all the values in between. Or use `group` with some fake values and the mask as a grouping coord. Or some NumPy function. Would need to test which is most performant for up to about 10 million pixels (though we should also try up to 100 million, I think).

src/scippnexus/nxdata.py
```diff
@@ -579,6 +579,60 @@ def _maybe_event_field(name: str, child: Union[Field, Group]):
...
+
+        masks = {}
+        for bit in range(number_of_bits_in_dtype):
+            steps_to_first_bit = number_of_bits_in_dtype - bit - 1
+            # Check if the mask associated with the current `bit` is present
+            mask_is_present = (masks_present >> steps_to_first_bit) % 2
+            if mask_is_present:
+                name = bit_to_mask_name.get(bit, f'pixel_mask_{bit}') + (
+                    f'_{suffix}' if suffix else ''
+                )
```
SimonHeybrock, line 628: In the fallback case, it is maybe a bit confusing that setting bit `N` in `pixel_mask_M` leads to a mask named `pixel_mask_N_M`?

Should we drop the "pixel_mask" bit from the name and simply name it `bit_N_M` or `bitN_M`? Or, going kind of with the standard: `undefined_bitN_M`?

tests/nxdata_test.py
```diff
@@ -674,3 +674,30 @@ def test_scalar_signal_without_unit_works(h5root):
     snx.create_field(data, 'xx', da.coords['xx'])
     group = snx.Group(data, definitions=snx.base_definitions())
     assert sc.identical(group[...], da)
+
+
+@pytest.mark.parametrize('bits', (32, 64))
```
SimonHeybrock, line 679: Also check bool, char, int16 pixel_masks?

src/scippnexus/nxdata.py
```diff
@@ -579,6 +579,60 @@ def _maybe_event_field(name: str, child: Union[Field, Group]):
...
+            key.lstrip('pixel_mask'): dg.pop(key)
+            # tuple because we are going to change the dict over the iteration
+            for key in tuple(dg.keys())
+            if key.startswith('pixel_mask')
+        }
+
+        array_or_dataset = super().assemble(dg)
+
+        for suffix, bitmask in bitmasks.items():
+            masks = self.transform_bitmask_to_dict_of_masks(bitmask, suffix)
```
SimonHeybrock, line 593: Won't you get a double `__` in the result?

tests/nxdata_test.py
```diff
@@ -674,3 +674,30 @@ def test_scalar_signal_without_unit_works(h5root):
...
+    )
+    masks = snx.NXdetector.transform_bitmask_to_dict_of_masks(bitmask)
+
+    assert np.all(masks.get('gap').values == np.array([1, 0, 0, 0, 0]))
+    assert np.all(masks.get('dead').values == np.array([0, 1, 0, 0, 0]))
+    assert np.all(masks.get('under_responding').values == np.array([0, 0, 1, 0, 0]))
+    assert np.all(masks.get('noisy').values == np.array([0, 0, 0, 0, 1]))
+    assert 'virtual_pixel' not in masks
+    assert 'user_defined_pixel' not in masks
+
```
SimonHeybrock, line 694: Need more tests with actual `NXdetector`, should probably go in to `nxdetector_test.py`

src/scippnexus/nxdata.py
jokasimr: I like `undefined_bitN_M`. It makes it clear the first number is the bit (and hopefully the user can then tell the second is the `M` in `pixel_maskM`).

src/scippnexus/nxdata.py
jokasimr: Yes you're right. I thought the naming from the Nexus standard was `pixel_maskN` instead of `pixel_mask_N`

tests/nxdata_test.py
jokasimr: How do I create a `scipp.array` with `int16` or `char`?

SimonHeybrock: You cannot, but if you use the underlying `h5py` directly you can write a dataset with that dtype from a NumPy array.

Admittedly, you will not see that dtype when this function is called, so this test only makes sense in the context of writing an NXdetector (see comment below on testing in that context)

src/scippnexus/nxdata.py
```diff
@@ -627,7 +627,7 @@ def transform_bitmask_to_dict_of_masks(bitmask: sc.Variable, suffix: str = None)
             # Check if the mask associated with the current `bit` is present
             mask_is_present = (masks_present >> steps_to_first_bit) % 2
             if mask_is_present:
-                name = bit_to_mask_name.get(bit, f'pixel_mask_{bit}') + (
+                name = bit_to_mask_name.get(bit, f'undefined_bit{bit}') + (
                     suffix if suffix is not None else ''
                 )
```
SimonHeybrock, line 632: You could simply set `suffix: str = ''`, the current type hint is wrong anyway, as it is.

tests/nxdata_test.py
jokasimr: I don't understand the examples in `nxdetector_test.py`. How should I go about creating a mock `NXdetector` with `pixel_mask`s to test the logic in `NXdetector.assemble`?

tests/nxdata_test.py
jokasimr: `snx.NXdetector.transform_bitmask_to_dict_of_masks(bitmask)` currently assumes the `bitmask` argument to be a `sc.Variable`.

If an `int16` occurs in a dataset then it will be casted when it is read by scippnexus, right?
If a pixel mask that was originally `int16` is casted to a `int32` variable then `snx.NXdetector.transform_bitmask_to_dict_of_masks` will not be able to interpret it correctly.

Is there some way of knowing what dtype a particular field (such as `pixel_mask`) had before it was converted to a variable and stored in the datagroup passed to `.assemble()`?

tests/nxdata_test.py
SimonHeybrock: Something like this would be a good start: https://github.com/scipp/scippnexus/blob/0e906ccb1dc36cb621d9903d5252b00d8fc467ee/tests/nxdetector_test.py#L126-L138

tests/nxdata_test.py
jokasimr: Because `transform_bitmask_to_dict_of_masks` assumes the most significant bit is the `gap` mask, the second most significant bit is the `dead` mask etc.. If the `int32` value was actually a `int16` bitmask originally then the 16 first bits are `0` and the bit associated with the `gap` mask is now at bit 16 in the `int32` data.
Unless there's something I've misunderstood.

SimonHeybrock: I thought `gap` = bit 0 = least significant bit?

jokasimr: This could be fixed by passing the number of bits in the bitmask explicitly.
That would actually be quite good because it would remove the special case we have now for `boolean` data.
But to do that we need to know the original dtype the `pixel_mask` field had in the file.

jokasimr: > I thought `gap` = bit 0 = least significant bit?

Oh.. Yes you're probably right. That simplifies things.

tests/nxdetector_test.py
```diff
@@ -687,3 +687,95 @@ def test_falls_back_to_hdf5_dim_labels_given_partially_axes(h5root):
...
+
+def test_pixel_masks_adds_suffix():
+    bitmask = sc.array(
+        dims=['detector_pixel'],
+        values=(1 << np.array([0, 1, 2, -1, 4])),
+        dtype='int32',
+    )
+    masks = snx.NXdetector.transform_bitmask_to_dict_of_masks(bitmask, '_test')
+    assert len(masks) == 4
+    assert all(k.endswith('_test') for k in masks.keys())
```
SimonHeybrock, line 728: Maybe we should use integer suffixes as in the standard instead of `test`?

tests/nxdetector_test.py
```diff
@@ -687,3 +687,95 @@ def test_falls_back_to_hdf5_dim_labels_given_partially_axes(h5root):
...
+        dtype='int64',
+    )
+    masks = snx.NXdetector.transform_bitmask_to_dict_of_masks(bitmask)
+
+    assert np.all(masks.get('gap').values == np.array([1, 0, 0, 0, 0]))
+    # A 'boolean' bitmask can only define one mask
+    if dtype != 'bool':
+        assert np.all(masks.get('dead').values == np.array([0, 1, 0, 0, 0]))
+        assert np.all(masks.get('under_responding').values == np.array([0, 0, 1, 0, 0]))
+        assert np.all(masks.get('noisy').values == np.array([0, 0, 0, 0, 1]))
```
SimonHeybrock, line 715: Use `scipp.identical` or `scipp.testing.assert_identical` fur such assertions. That will also ensure unit and dtype are as expected.

tests/nxdetector_test.py
```diff
@@ -687,3 +687,95 @@ def test_falls_back_to_hdf5_dim_labels_given_partially_axes(h5root):
...
+    snx.create_field(detector, 'data', da.data)
+    snx.create_field(detector, 'pixel_mask', bitmask)
+    snx.create_field(detector, 'pixel_mask_2', bitmask)
+    detector.attrs['axes'] = ['xx', '.']
+    detector = make_group(detector)
+    da = detector[...]
+    assert np.allclose(da.masks.get('gap').values, np.array([[1, 0], [0, 0]]))
+    assert np.allclose(da.masks.get('dead').values, np.array([[0, 1], [0, 0]]))
+    assert np.allclose(da.masks.get('gap_2').values, np.array([[1, 0], [0, 0]]))
+    assert np.allclose(da.masks.get('dead_2').values, np.array([[0, 1], [0, 0]]))
```
SimonHeybrock, line 760: `allclose`? See above about using Scipp's `identical` or `assert_identical`.

src/scippnexus/nxdata.py
```diff
@@ -579,6 +579,57 @@ def _maybe_event_field(name: str, child: Union[Field, Group]):
...
+    def transform_bitmask_to_dict_of_masks(bitmask: sc.Variable, suffix: str = ''):
+        bit_to_mask_name = {
+            0: 'gap',
+            1: 'dead',
+            2: 'under_responding',
+            3: 'virtual_pixel',
+            4: 'noisy',
+            6: 'pixel_is_part_of_a_cluster_of_problematic_pixels',
+            8: 'user_defined_mask',
+            31: 'virtual_pixel',
```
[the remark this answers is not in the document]
jokasimr: That makes sense. But there is also some value in following the standard closely to avoid ambiguity. I'm leaning towards using the names in the nexus standard, but I'm open for changing them if you think that's better.

src/scippnexus/nxdata.py
```diff
@@ -579,6 +579,57 @@ def _maybe_event_field(name: str, child: Union[Field, Group]):
...
+        }
+
+        number_of_bits_in_dtype = 8 * bitmask.values.dtype.itemsize
+
+        # Bitwise indicator of what masks are present
+        masks_present = np.bitwise_or.reduce(bitmask.values.ravel())
```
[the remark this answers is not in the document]
jokasimr: Because we want to reduce over all dimensions of the array to a single scalar. By default the reduction is done along `axis=0`, and for non-1d input it will return a non-scalar output.

src/scippnexus/nxdata.py
```diff
@@ -579,6 +579,57 @@ def _maybe_event_field(name: str, child: Union[Field, Group]):
...
+
+        masks = {}
+        for bit in range(number_of_bits_in_dtype):
+            # Check if the mask associated with the current `bit` is present
+            if (masks_present >> bit) % 2:
```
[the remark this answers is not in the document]
jokasimr: Might be safer!

src/scippnexus/nxdata.py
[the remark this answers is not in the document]
SimonHeybrock: The difference is that in Scipp you can also have non-pixel masks (such as a TOF or wavelength mask), whereas the names in the standard are specifically in the context of the `pixel_mask` field. `gap` may not make much sense to a user who never looks at the file?

src/scippnexus/nxdata.py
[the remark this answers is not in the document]
SimonHeybrock: Oh, I thought it reduces all dims? So this is unlike, e.g., `np.sum()`?

tests/nxdetector_test.py
```diff
@@ -687,3 +687,95 @@ def test_falls_back_to_hdf5_dim_labels_given_partially_axes(h5root):
...
+            # Set mask 0, 1, 2, and 4 but not 3.
+            1 << np.array([0, 1, 2, -1, 4])
+            if dtype != 'bool'
+            else
+            # Set mask 0 only.
+            np.array([1, 0, 0, 0, 0])
+        )
+        .astype(dtype)
+        .astype('int64'),
+        dtype='int64',
```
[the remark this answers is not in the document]
jokasimr: I was thinking that this mimics what happens when a mask is read from a nexus file.
* When the mask is read from the nexus file it is a Numpy array. (is this correct?)
* Then it is cast to a dtype that scipp supports (`bool` or `int32` or `int64`).
* Then a scipp variable is created from it.

If we don't cast 'int8' or 'int16' before creating the variable we get `RuntimeError: Unsupported numpy dtype: int16`.

What change are you suggesting?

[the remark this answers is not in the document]
SimonHeybrock: > Then it is cast to a dtype that scipp supports (bool or int32 or int64).

Yes, so why do you only test `int64`? Why do you even have a parametrized test then?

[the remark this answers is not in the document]
jokasimr: > Why do you even have a parametrized test then?

To make sure the function works in case the pixel-mask arrays loaded from the nexus file have different datatypes.

I can change it to only test dtypes supported by scipp if you think that makes more sense.

src/scippnexus/nxdata.py
```diff
@@ -603,13 +603,13 @@ def assemble(self, dg: sc.DataGroup) -> Union[sc.DataArray, sc.Dataset]:
...
         bit_to_mask_name = {
-            0: 'gap',
-            1: 'dead',
-            2: 'under_responding',
-            3: 'over_responding',
-            4: 'noisy',
+            0: 'gap_pixel',
+            1: 'dead_pixel',
+            2: 'under_responding_pixel',
+            3: 'over_responding_pixel',
```
SimonHeybrock, line 609:
```suggestion
            2: 'underresponding_pixel',
            3: 'overresponding_pixel',
```
I think these are actually single words?

tests/nxdetector_test.py
```diff
@@ -781,29 +725,36 @@ def test_pixel_masks_reads_expected_fields(h5root):
...
             ),
-            values=[[0, 1], [0, 0]],
+            values=[[1, 0], [0, 0]],
             dtype='bool',
         ),
     )
+
+    # A 'boolean' bitmask can only define one mask
+    if dtype == 'bool':
+        print(bitmask, list(da.masks.keys()))
```
SimonHeybrock, line 741:
```suggestion
```

src/scippnexus/nxdata.py
jokasimr: In the nexus standard they are separated by space. Do you think we should join them anyway?

tests/nxdetector_test.py
jokasimr: :sweat_smile:

src/scippnexus/nxdata.py
jokasimr: I'll join them. It does look better.

SimonHeybrock (approved):

PR #167, opened 2023-10-05
Save errors in create_field

### LINKED ISSUE #105: `NXobject.create_field` and `__setitem__` ignoring variances when passing `scipp.Variable`?
Check the code, and add a test.

jl-wynen:
Fixes #105

### DIFF
--- a/src/scippnexus/base.py
+++ b/src/scippnexus/base.py
@@ -241,7 +241,7 @@ class Group(Mapping):
                 if (
                     isinstance(values, Field)
                     and isinstance(errors, Field)
-                    and values.unit == errors.unit
+                    and (values.unit == errors.unit or errors.unit is None)
                     and values.dataset.shape == errors.dataset.shape
                 ):
                     values.errors = errors.dataset
--- a/src/scippnexus/base.py
+++ b/src/scippnexus/base.py
@@ -390,7 +390,14 @@ class Group(Mapping):
         """Create a child dataset with given name and value.
 
         Note that due to the caching mechanisms in this class, reading the group
-        or its children may not reflect the changes made by this method."""
+        or its children may not reflect the changes made by this method.
+
+        Returns
+        -------
+        :
+            The created dataset of the values.
+            If errors are written to the file, their dataset is not returned.
+        """
         return create_field(self._group, key, value)
 
     def create_class(self, name: str, class_name: str) -> Group:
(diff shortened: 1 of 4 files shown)


src/scippnexus/base.py
```diff
@@ -424,23 +431,46 @@ def shape(self) -> Tuple[int, ...]:
...
+        values, errors, attrs = _create_field_params_number(data)
+
+    if isinstance(data, sc.Variable) and data.unit:
+        attrs['units'] = str(data.unit)
+
+    values_dataset = group.create_dataset(name, data=values, **kwargs)
+    values_dataset.attrs.update(attrs)
+    if errors is not None:
+        errors_dataset = group.create_dataset(name + '_errors', data=errors, **kwargs)
+        errors_dataset.attrs.update(attrs)
```
SimonHeybrock, line 471: I am not sure the `units` attribute (or any others) should we written here?

SimonHeybrock: On the other hand... I can't see any harm, even if the standard does not require it?

SimonHeybrock (approved):

src/scippnexus/base.py
jl-wynen: I think the standard is silent on this. I did it just in case they are useful. But I'm fine with not writing them.

SimonHeybrock: Can you check with Tobias? If the standard does not specify them, maybe we should not write them either. Otherwise we may end up in situations where our reading code (which uses this for testing) only works with files that where written using ScippNexus.

src/scippnexus/base.py
```diff
@@ -235,13 +235,14 @@ def _make_child(obj: Union[H5Dataset, H5Group]) -> Union[Field, Group]:
         items = {k: v for k, v in items.items() if not k.startswith('cue_')}
         for suffix in ('_errors', '_error'):
             field_with_errors = [name for name in items if f'{name}{suffix}' in items]
+            print(suffix, field_with_errors)
```
SimonHeybrock, line 238:
```suggestion
```

jl-wynen: Oops. We should use https://pypi.org/project/flake8-print/. Or switch to Ruff like Scitacean: https://docs.astral.sh/ruff/rules/#flake8-print-t20

tests/nexus_test.py
```diff
@@ -434,6 +434,24 @@ def test_errors_read_as_variances(h5root):
     assert np.array_equal(dg['time'].variances, np.arange(5.0) ** 2)
 
 
+def test_does_not_require_unit_of_errors(h5root):
```
SimonHeybrock, line 437: With the latest change, is there a test that checks behavior for mismatching error units?

jl-wynen: Yes: https://github.com/scipp/scippnexus/blob/54642fe7660fad7811a749c4e8327ae9a6726afc/tests/nexus_test.py#L215

SimonHeybrock (approved):

PR #169, opened 2023-10-23
Update for next scipp

jl-wynen:


tests/nxdata_test.py
```diff
@@ -216,7 +216,7 @@ def test_auxiliary_signal_causes_load_as_dataset(h5root):
     snx.create_field(data, 'signal', da.data)
     snx.create_field(data, 'xx', da.coords['xx'])
     data = snx.Group(data, definitions=snx.base_definitions())
-    assert_identical(data[...], sc.Dataset({'signal': da.data, 'xx': da.coords['xx']}))
+    assert_identical(data[...], sc.DataGroup(signal=da.data, xx=da.coords['xx']))
```
jl-wynen, line 219: Honestly not sure what happened here. But in my local build, `data[...]` returns a datagroup. But not in CI.

SimonHeybrock: I do not know if NeXus guarantees that aux signals have the same number of dims, so maybe we have to consider moving to `DataGroup` anyway?

jl-wynen: Seems like this happens anyway which is why I had to fix the test. But I don't know what triggers it. I didn't change the implementation of `Group`

src/scippnexus/nxcylindrical_geometry.py
```diff
@@ -19,10 +19,13 @@ def _parse(
     face1_center = cylinders['vertex_index', 0]
     face1_edge = cylinders['vertex_index', 1]
     face2_center = cylinders['vertex_index', 2]
-    ds = sc.Dataset()
-    ds['face1_center'] = vertices[face1_center.values]
-    ds['face1_edge'] = vertices[face1_edge.values]
-    ds['face2_center'] = vertices[face2_center.values]
+    ds = sc.Dataset(
```
nvaytet, line 22: Can you also fix this in `v1`? It's showing up when using `load_nexus`.
I know it's deprecated but I still have to use it until we handle the transformation chains.

jl-wynen: Done. Unfortunately, there are no tests for v1. So I don't know for sure that it works.

tests/nxdata_test.py
SimonHeybrock: It probably means that there is an exception somewhere, so it falls back to loading as a datagroup. This likely indicates your broke something elsewhere. Do not change this test!

jl-wynen: It needs to change because signal and xx have different shapes

SimonHeybrock: Looking at the standard, I see now that the aux-signals are suppose to have the same shape as the original. That is, the existing test was a bit bad. Therefore:

- Change the shape of the aux signal stored in this test.
- Keep checking for `Dataset`.

SimonHeybrock (approved):

PR #170, opened 2023-10-25
Move to copier

SimonHeybrock:


.github/workflows/ci.yml
```diff
@@ -24,28 +35,23 @@ jobs:
...
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v3
-        with:
-          python-version: ${{ matrix.python }}
-      - run: python -m pip install --upgrade pip
-      - run: python -m pip install -r requirements/ci.txt
-      - run: tox -e ${{ matrix.tox }}
+        os: ['ubuntu-20.04']
+        python:
+          - version: '${{needs.formatting.outputs.min_python}}'
```
jl-wynen, line 44: So now this only runs with a single python version?

.github/workflows/release.yml
```diff
@@ -57,65 +64,53 @@ jobs:
...
     steps:
       - uses: actions/download-artifact@v3
-      - uses: conda-incubator/setup-miniconda@v2
+      - uses: mamba-org/setup-micromamba@v1
         with:
-          python-version: '3.10'
-      - run: conda install -c conda-forge --yes anaconda-client
-      - run: anaconda --token ${{ secrets.ANACONDATOKEN }} upload --user scipp --label main $(ls conda-package-*/*/*.tar.bz2)
-
-  manage-versions:
```
jl-wynen, line 85: Do we need to manually clean up the gh-pages branch?

docs/developer/getting-started.md
```diff
@@ -0,0 +1,91 @@
...
+tox -e linkcheck
+```
+````
+
+````{tab-item} Manually
+
+Build the documentation using
+
+```sh
+python -m sphinx -v -b html -d build/.doctrees docs build/html
```
jl-wynen, line 81: This still uses a `build` folder. But tox doesn't and the folder is no longer ignored by git.

.github/workflows/ci.yml
SimonHeybrock: Yes, I considered keeping multiple versions, but felt there was little benefit at this point.

.github/workflows/release.yml
SimonHeybrock: I think the settings in the deploy action in `docs.yml` might wipe that automatically?

jl-wynen: Looks like it. Let's keep an eye on it in the next release.

docs/developer/getting-started.md
SimonHeybrock: Good catch, I opened https://github.com/scipp/copier_template/issues/63.

jl-wynen (approved):
The branch protection rules need to be updated. Otherwise, lgtm

PR #172, opened 2023-10-26
Load NXdetector and NXmonitor as DataGroup

SimonHeybrock:
This avoids some conceptual problems and ambiguities. I think we might need to refine some details, e.g., some scalar datasets should probably made into coords, instead of being kept in the top-level data group.

This constitutes a breaking change for users. One could probably support both approaches side-by-side by having the old and new `NXdetector` and `NXmonitor` definitions, which can be specified when opening a file using `snx.File`. Should we do that, or is our user base small enough so we can move directly?

### DIFF
--- a/docs/user-guide/nexus-classes.ipynb
+++ b/docs/user-guide/nexus-classes.ipynb
@@ -107,7 +107,7 @@
     "\n",
     "Provides multi-dimensional labeled data.\n",
     "See the NeXus format [NXdata base class definition](https://manual.nexusformat.org/classes/base_classes/NXdata.html) for details.\n",
-    "Can be read as a data array using slicing syntax.\n",
+    "Can be read as a data array using positional indexing.\n",
     "\n",
     "Example:"
    ]
@@ -145,7 +145,7 @@
     "Provides data for a detector.\n",
     "See the NeXus format [NXdetector base class definition](https://manual.nexusformat.org/classes/base_classes/NXdetector.html) for details.\n",
     "`NXdetector` contains data and coords similar to `NXdata` as well as additional fields that are not readily inserted into a `scipp.DataArray`.\n",
-    "Therefore, `NXdetector` can be read as a `scipp.DataGroup` wrapping a `scipp.DataArray` using slicing syntax.\n",
+    "Therefore, `NXdetector` can be read as a `scipp.DataGroup` wrapping a `scipp.DataArray` using positional indexing.\n",
     "The \"signal\" field and associated coordinates are combined into a `scipp.DataArray` and the remaining fields are added as fields to the `scipp.DataGroup`.\n",
     "In the output, the `scipp.DataArray` has the same name as the \"signal\" dataset in the NeXus file.\n",
     "The underlying data may be dense data or event data.\n",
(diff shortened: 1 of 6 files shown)


src/scippnexus/nxdata.py
```diff
@@ -347,6 +352,42 @@ def assemble(
...
+        """
+        data_items = sc.DataGroup()
+        result = sc.DataGroup()
+        signals = [self._signal_name] + self._aux_signals
+        for name, value in dg.items():
+            # We need the shape *before* slicing to determine dims, so we get the
+            # field from the group for the conditional.
+            if name not in signals and (
+                isinstance(value, (sc.DataArray, sc.DataGroup))
+                or self._children[name].shape == ()
```
SimonHeybrock, line 378: I'll look into a better option, converting to draft for now.

jl-wynen:
> Should we do that, or is our user base small enough so we can move directly?

Since the new approach would become the default, the users would need to take action immediately anyway (unless we do 2 step approach like migration to plopp). And converting the new data groups into data arrays isn't much more complicated than switching to the old definitions. Esp. because users probably are not used to application definitions.

src/scippnexus/nxdata.py
```diff
@@ -347,6 +354,46 @@ def assemble(
...
+        for name, value in dg.items():
+            if name in forward:
+                data_items[name] = value
+            else:
+                result[name] = value
+        assembled_nxdata = self._assemble_as_data(data_items)
+        result.update(
+            {self._signal_name: assembled_nxdata}
+            if isinstance(assembled_nxdata, sc.DataArray)
+            else assembled_nxdata
```
jl-wynen, line 393: This unpacks datasets into `result`. Is this on purpose?

tests/nxdetector_test.py
jl-wynen, line 1: If I'm not mistaken, there is a discrepancy between loading a detector and the contained data: `f['detector'][()]['data']` contains all coords but `f['detector']['data'][()]` only contains the axes, right?
Is this a problem?

docs/user-guide/nexus-classes.ipynb
```diff
@@ -142,7 +144,10 @@
     "\n",
     "Provides data for a detector.\n",
     "See the NeXus format [NXdetector base class definition](https://manual.nexusformat.org/classes/base_classes/NXdetector.html) for details.\n",
-    "Can be read as a data array using slicing syntax.\n",
+    "`NXdetector` contains data and coords similar to `NXdata` as well as additional fields that are not readily inserted into a `scipp.DataArray`.\n",
+    "Therefore, `NXdetector` can be read as a `scipp.DataGroup` wrapping a `scipp.DataArray` using slicing syntax.\n",
```
jl-wynen, line 148: What do you mean by 'using slicing syntax'?

src/scippnexus/nxdata.py
SimonHeybrock: Yes, there is no need for keeping the dataset. Result is like this:
image

tests/nxdetector_test.py
SimonHeybrock: This bothered me a bit as well. I think being able to load an individual dataset makes more sense, in particular since before loading the structure we assemble into is far from obvious.

src/scippnexus/nxdata.py
jl-wynen: :+1:

docs/user-guide/nexus-classes.ipynb
SimonHeybrock: Changed to "positional indexing" everywhere.

jl-wynen (approved):

PR #173, opened 2023-11-01
Add helper to compute positions from depends_on

### LINKED ISSUE #160: Add functions or module for resolving `depends_on` chains
Also check if there are existing pieces for executing depends_on chains that need refactoring.

SimonHeybrock:
Fixes #160.

### DIFF
--- a/src/scippnexus/nxtransformations.py
+++ b/src/scippnexus/nxtransformations.py
@@ -205,6 +205,11 @@ class TransformationChainResolver:
     class follows the paths and resolves the chain of transformations.
     """
 
+    class ChainError(KeyError):
+        """Raised when a transformation chain cannot be resolved."""
+
+        pass
+
     def __init__(self, stack: List[sc.DataGroup]):
         self._stack = stack
 
@@ -215,7 +220,9 @@ class TransformationChainResolver:
     @property
     def parent(self) -> TransformationChainResolver:
         if len(self._stack) == 1:
-            raise KeyError("Transformation depends on node beyond root")
+            raise TransformationChainResolver.ChainError(
+                "Transformation depends on node beyond root"
+            )
         return TransformationChainResolver(self._stack[:-1])
 
     @property
@@ -368,7 +375,7 @@ def _with_positions(
             out[store_position] = transform * sc.vector([0, 0, 0], unit='m')
             if store_transform is not None:
                 out[store_transform] = transform
-        except KeyError:
+        except TransformationChainResolver.ChainError:
             pass
     for name, value in dg.items():
         if isinstance(value, sc.DataGroup):
--- a/src/scippnexus/nxtransformations.py
+++ b/src/scippnexus/nxtransformations.py
@@ -235,11 +235,10 @@ class TransformationChainResolver:
         depends_on = self.value.get('depends_on')
         if depends_on is None:
             return None
-        origin = sc.vector([0, 0, 0], unit='m')
         # Note that transformations have to be applied in "reverse" order, i.e.,
-        # simply taking math.prod(chain) * origin would be wrong, even if we could
+        # simply taking math.prod(chain) would be wrong, even if we could
         # ignore potential time-dependence.
-        return combine_transformations(self.get_chain(depends_on)) * origin
+        return combine_transformations(self.get_chain(depends_on))
 
     def get_chain(self, depends_on: str) -> List[Union[sc.DataArray, sc.Variable]]:
         if depends_on == '.':
--- a/src/scippnexus/nxtransformations.py
+++ b/src/scippnexus/nxtransformations.py
@@ -135,6 +135,12 @@ def _smaller_unit(a, b):
 def combine_transformations(
     chain: List[Union[sc.DataArray, sc.Variable]]
 ) -> Union[sc.DataArray, sc.Variable]:
+    """
+    Take the product of a chain of transformations, handling potentially mismatching
+    time-dependence.
+
+    Time-dependent transformations are interpolated to a common time-coordinate.
+    """
     total_transform = None
     for transform in chain:
         if total_transform is None:
@@ -165,7 +171,7 @@ def combine_transformations(
         times = [da.coords['time'][0] for da in time_dependent]
         latest_log_start = sc.reduce(times).max()
         return total_transform['time', latest_log_start:].copy()
-    return 1 if total_transform is None else total_transform
+    return sc.scalar(1) if total_transform is None else total_transform
 
 
 def maybe_transformation(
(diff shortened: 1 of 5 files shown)


src/scippnexus/nxtransformations.py
```diff
@@ -228,7 +184,143 @@ def maybe_transformation(
...
+                depends_on = attr.value
+            # If transform is time-dependent then we keep it is a DataArray, otherwise
+            # we convert it to a Variable.
+            transform = transform if transform.coords else transform.data
+        if transform.dtype in (sc.DType.translation3, sc.DType.affine_transform3):
+            transform = transform.to(unit='m', copy=False)
+        return [transform] + node.parent.get_chain(depends_on)
+
+
+def compute_positions(dg: sc.DataGroup, *, store_as: str = 'position') -> sc.DataGroup:
```
jl-wynen, line 261: I'm not sure this is a good interface. It resolves all transforms in the data group. This may not be what a user wants in some cases. And most importantly, there may be chains that lead outside the provided group while the one that the user cares about is fully specified. So the operation would fail even though it doesn't have to.

tests/nxtransformations_test.py
```diff
@@ -432,3 +432,184 @@ def test_slice_transformations(h5root):
...
+
+
+def test_TransformationChainResolver_path_handling():
+    tree = TransformationChainResolver([{'a': {'b': {'c': 1}}}])
+    assert tree['a']['b']['c'].value == 1
+    assert tree['a/b/c'].value == 1
+    assert tree['/a/b/c'].value == 1
+    assert tree['a']['../a/b/c'].value == 1
+    assert tree['a/b']['../../a/b/c'].value == 1
+    assert tree['a/b']['./c'].value == 1
```
jl-wynen, line 444: Missing cases for failures
- non existent path segment 
- path that leads outside the tree

jl-wynen: What about trailing slashes?

tests/nxtransformations_test.py
jl-wynen, line 1: Do we need to worry about cycles in depends-on chains? Or is it ok to raise a recursion limit error because this would be caused by a broken file?

tests/nxtransformations_test.py
```diff
@@ -432,3 +432,184 @@ def test_slice_transformations(h5root):
...
+                'child1': {'grandchild1': transform1},
+                'child2': transform2,
+            }
+        ]
+    )
+    expected = transform2.data * transform1.data * origin
+    assert sc.identical(tree.resolve_depends_on(), expected)
+
+
+def test_compute_positions(h5root):
```
jl-wynen, line 515: Wouldn't it be easier to create the data group directly? Or do you also want to test that the loaded group is compatible with `compute_positions`?

src/scippnexus/nxtransformations.py
```diff
@@ -228,7 +184,143 @@ def maybe_transformation(
...
+    resolver = TransformationChainResolver([dg])
+    return _with_positions(dg, store_as=store_as, resolver=resolver)
+
+
+def _zip_pixel_offsets(da: sc.DataArray) -> sc.Variable:
+    zero = sc.scalar(0.0, unit=da.coords['x_pixel_offset'].unit)
+    return sc.spatial.as_vectors(
+        da.coords['x_pixel_offset'],
+        da.coords.get('y_pixel_offset', zero),
+        da.coords.get('z_pixel_offset', zero),
```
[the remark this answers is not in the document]
SimonHeybrock: In practice I have never seen only `x`, but we may get only `x` and `y` for flat 2D panels, I think.

src/scippnexus/nxtransformations.py
SimonHeybrock: > This may not be what a user wants in some cases

Sounds hypothetical, in general most or all positions will be required, I think?

tests/nxtransformations_test.py
SimonHeybrock: The latter.

src/scippnexus/nxtransformations.py
jl-wynen: Yes, in general. But this interface means that it is impossible to resolve *any* chains unless the relevant portions of the file are fully intact. As an example, let's say we have a file with both a monitor and a detector in an instrument group. The detector's chain leads it out of its own group and into the instrument. But the monitor's chain is broken. Now we cannot resolve the transformations of the detector even though it is intact.
Given all the things I have seen with nexus files, I don't think this is a far fetched scenario that people will face at least in commissioning.

src/scippnexus/nxtransformations.py
```diff
@@ -228,7 +184,143 @@ def maybe_transformation(
     transformation fields.
     """
     if (transformation_type := obj.attrs.get('transformation_type')) is not None:
-        from .nxtransformations import Transformation
-
         return Transformation(obj).make_transformation(value, transformation_type, sel)
     return value
+
+
+class TransformationChainResolver:
```
jl-wynen, line 191: Could we do away with this class if we implemented https://github.com/scipp/scipp/issues/3143?

SimonHeybrock: No, because `DataGroup` does not have parent references.

SimonHeybrock:
Converting to draft, since some of this is most definitely wrong: I think I was badly mislead by the NeXus documentation.

src/scippnexus/nxtransformations.py
jl-wynen: We would have to resolve all paths in the data group that the user passes in. But I gues this would still require keeping track of the current path along the lines of
```python
base_dg[path_up_to_the_current_element / depends_on]
```
I'm just hoping to avoid implementing the traversal twice.

SimonHeybrock: > We would have to resolve all paths in the data group that the user passes in.

How? In NeXus you may have a relative path deeply nested in some subgroup. I don't see how `DataGroup` could ever help with that, unless we add parent references.

jl-wynen: In my example, `path_up_to_the_current_element` is the path to that deeply nested element. So combining it with the `depends_on` attr yields a path relative to the root data group.

SimonHeybrock: You would still need a wrapper class that does the traversal and path merging. And you would still need to check if `depends_on` is absolute or relative. So I am not sure you would gain much?

src/scippnexus/nxtransformations.py
```diff
@@ -228,7 +184,143 @@ def maybe_transformation(
...
+    def parent(self) -> TransformationChainResolver:
+        if len(self.path) == 1:
+            raise TransformationError("Transformation depends on node beyond root")
+        return TransformationChainResolver(self.path[:-1])
+
+    @property
+    def value(self) -> sc.DataGroup:
+        return self.path[-1]
+
+    def __getitem__(self, path: str) -> TransformationChainResolver:
```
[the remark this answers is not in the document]
SimonHeybrock: Renamed to 'stack'

tests/nxtransformations_test.py
SimonHeybrock: `depends_on` always refers to a Nexus field, never to a Nexus class, so I don't think trailing slashes should be valid?

tests/nxtransformations_test.py
SimonHeybrock: I'd say raising is fine.

src/scippnexus/nxtransformations.py
```diff
@@ -348,10 +363,13 @@ def _with_positions(
...
-        transform = resolver.resolve_depends_on()
-        out[store_position] = transform * sc.vector([0, 0, 0], unit='m')
-        if store_transform is not None:
-            out[store_transform] = transform
+        try:
+            transform = resolver.resolve_depends_on()
+            out[store_position] = transform * sc.vector([0, 0, 0], unit='m')
+            if store_transform is not None:
+                out[store_transform] = transform
+        except KeyError:
```
jl-wynen, line 371: Where is this error raised and why is it ok to ignore? Is it about this `raise KeyError("Transformation depends on node beyond root")`? If so, I'd prefer a check of the error message so as not to ignore other errors. There might be no way to raise KeyError now but that can change.

SimonHeybrock: Yes, this was in reply to your comment on the earlier behavior. Catching more specific now.

jl-wynen (approved):

src/scippnexus/nxtransformations.py
SimonHeybrock: @jl-wynen Of course this broke immediately with a new real file: After the latest change `compute_positions` fails if `depends_on` points to, e.g., a non-existent field in the current folder. I think this should be handled no different from pointing to a parent, or should it?

jl-wynen: I guess so.

Both cases might be hard to debug. Can you add a warning (togglable with a function arg) that explains what element's positions are not computed and why?

PR #185, opened 2023-12-14
Convert offsets to float64

nvaytet:
When trying to load a file and compute the transformations using `compute_positions`, applying a transformation fails if the `[xyz]_pixel_offset` are `float32`, because we cannot make vectors with `float32` dtype in scipp.

With this change, we try to perform a conversion on-the-fly to float64 before sending to `sc.spatial.as_vectors()`


jl-wynen:
Don't you want to implement https://github.com/scipp/scipp/issues/3329 instead?

SimonHeybrock:
Agree, this should be closed in favor of fixing scipp.

PR #189, opened 2024-01-17
Experiment with lock to avoid deadlocks

SimonHeybrock:
Related to #188. Adding a lock seems to bypass the issue.

This is not ready:
- If we go this route, we should add a flag to enable/disable the behavior.
- Should we lock in `__init__`, i.e., prevent file creation? Right now only the context-manager use is "fixed".

### DIFF
--- a/src/scippnexus/file.py
+++ b/src/scippnexus/file.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 # @author Simon Heybrock
+import threading
 from contextlib import AbstractContextManager
 from typing import Mapping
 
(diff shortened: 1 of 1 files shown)


src/scippnexus/file.py
```diff
@@ -12,6 +13,8 @@
 
 
 class File(AbstractContextManager, Group):
+    _lock = threading.Lock()
```
nvaytet, line 16: > Should we lock in __init__, i.e., prevent file creation? Right now only the context-manager use is "fixed".

How does it work when you don't use a context manager also for reading files?
In the docs, we have [examples](https://scipp.github.io/scippnexus/user-guide/quick-start-guide.html) of interactively working with an open file...

nvaytet (commented):
Really great if the solution is that simple :+1:

SimonHeybrock:
> Really great if the solution is that simple 👍

It is not a solution, it is a workaround, imho.

src/scippnexus/file.py
SimonHeybrock: It will work as before, as long as you only create a single `snx.File`.

SimonHeybrock:
Superseded by #190.

PR #192, opened 2024-01-26
Use pytest importlib import mode

jl-wynen:
Part of https://github.com/scipp/scipp/issues/3344

I tried converting `externalfile` into a fixture so that tests don't need the marker, i.e., something along the lines of
```python
# in conftest.py
@pytest.mark.fixture(scope='session')
def externalfile():
    from externalfile import ExternelFile
    return ExternalFile()

# in test file
def test_foo(externalfile):
    path = externalfile.get_path(name)
    ...
```

Where `ExternalFile` is a simple class that holds the pooch inventory. Unfortunately, I didn't find a way to skip those tests via a marker. So I think we would have to move to a dedicated command line option, see the setup in Scitacean: https://github.com/SciCatProject/scitacean/blob/main/tests/conftest.py

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ minversion = "7.0"
 addopts = """
 --strict-config
 --strict-markers
+--import-mode=importlib
 -ra
 -v
 -m 'not externalfile'
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,8 @@
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
 import os
+import sys
+from pathlib import Path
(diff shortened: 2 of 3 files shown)


SimonHeybrock:
I think `externalfile` might be used only by a single test file. Can we simply move the code there?

jl-wynen:
We could. But do you want to keep it that way when we have more tests / test files?

SimonHeybrock:
Should we consider making this data part of the package, instead of just the tests?

jl-wynen:
Just so we slightly simplify the import in the tests? I would say no, because it is not useful to package users.

SimonHeybrock (approved):

PR #194, opened 2024-02-09
feat: label based indexing for nxdata

### LINKED ISSUE #191: Support label-based indexing
We have always had in mind to implement [label-based indexing like Scipp](https://scipp.github.io/user-guide/slicing.html#Label-based-indexing). This is required in particular when loading multiple groups at once, where positional indexing is often not meaningful.

In fact, we may consider the current behavior a bug, or at least a discrepancy to how `scipp.DataGroup` behaves: When loading a tree of groups, the dimension length may be ambiguous (indicated by `None`). ScippNexus nevertheless accepts positional indices. Therefore, we get inconsistent behavior:

```python
with snx.File(name) as f:
    dg1 = f['time', 10:20]  # may return some zero-length subgroups, if indices out of range
    dg2 = f[()]['time', 10:20]  # raises if sizes is {'time': None, ...}
```

See scipp/scipp#3371.

jokasimr:
Fixes #191


src/scippnexus/nxdata.py
```diff
@@ -329,6 +334,26 @@ def index_child(self, child: Union[Field, Group], sel: ScippIndex) -> ScippIndex
...
         return child[child_sel]
 
+    def read_children(self, sel: ScippIndex) -> sc.DataGroup:
+        if is_label_based_index(sel):
+            coord, index = sel
+            if coord not in self._children:
+                raise ValueError(f'Coordinate {coord} was not found on the object')
+            child = self._children[coord][()]
+            if len(child.dims) != 1:
+                raise ValueError('The indexed coordinate must be 1-d')
```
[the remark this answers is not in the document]
jokasimr: I think it's probably fine to allow it, like you said it's unlikely to be an issue because the values has to be 1d, sorted etc for it to work, and if they are it's not really a problem anyway.
What is typically the name of the values dataset?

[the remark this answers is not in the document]
SimonHeybrock: Look at the fields set up by the NXdata init. There are some fields that indicate which fields are signals (there may be multiple).

SimonHeybrock:
Can you add the equivalent of https://github.com/scipp/scippnexus/blob/58ffbbe67e90bf3341a7589e89e00e36859ba712/tests/nxdetector_test.py#L279-L310, but with label-based indexing instead of positional indexing? Note that this is a special case because the `event_time_zero` coord is typically in a subgroup that is handled as a special "event field".

jokasimr:
I'm unsure about the `NXlog` case. Perhaps we should talk about that in person.

src/scippnexus/nxdata.py
```diff
@@ -487,11 +496,16 @@ def __init__(self, attrs: Dict[str, Any], children: Dict[str, Union[Field, Group
     def read_children(self, sel: ScippIndex) -> sc.DataGroup:
         # Sublogs have distinct time axes (with a different length). Must disable
         # positional indexing.
-        if self._sublogs and ('time' in _to_canonical_select(list(self.sizes), sel)):
+        if (
+            not is_label_index(sel)
+            and self._sublogs
+            and ('time' in _to_canonical_select(list(self.sizes), sel))
+        ):
```
SimonHeybrock, line 503: Have you actually tried this? From looking at the code it looks broken to me, as there is just a single label-based -> positional conversion in `read_children` of `NXdata`. But for NXlog like in this case it would need to be done separately for each sublog.

Maybe sublogs need to be read independently? If it gets too complicated, postpone support for this particular case for now (but add a test).

src/scippnexus/_common.py
```diff
@@ -106,6 +106,19 @@ def to_plain_index(dims: List[str], select: ScippIndex) -> Union[int, slice, tup
     return tuple(index)
 
 
+def is_label_index(s):
+    if isinstance(s, dict):
+        return all(map(is_label_index, s.items()))
```
SimonHeybrock, line 111: Not sure why it has to be all. One index may be positional, but another may be label-based.

src/scippnexus/nxdata.py
jokasimr: No not really. There's one test that involves slicing NxLog and that's the only time I've tried this. I don't precisely understand what should be done in the NxLog case. I think we will figure it out if we talk a bit about it and you explain a bit more how you envision it working.

Do you know of a good example nexus file that has NxLog data?

src/scippnexus/_common.py
jokasimr: Yeah, it's not really well defined. This function is only used in the NxLog case and since positional indexing is not allowed in that case the reasoning was that this function should only returns true if all slices are label based.

jokasimr: It should probably be renamed to `all_are_label_based_indices`

src/scippnexus/nxdata.py
SimonHeybrock: I suggest you do not handle this in this PR (but make sure it still raises). I have some data, but let us keep things simple here.

jokasimr: Okay do you mean don't handle `NxLog` at all in this PR, or not handle sublogs?

SimonHeybrock: Do not handle sublogs.

src/scippnexus/base.py
```diff
@@ -140,6 +141,30 @@ def assemble(
...
         return dg
 
+    def _convert_index_to_positional(self, sel):
+        if isinstance(sel, dict):
+            return dict((self._convert_index_to_positional(s) for s in sel.items()))
+        if not (isinstance(sel, tuple) and len(sel) > 1):
+            return sel
+        coord, index = sel
+        if coord not in self._children:
+            return sel
```
SimonHeybrock, line 151: I think we should actually raise if `sel` is a label but `coord` is not found. Otherwise we get confusing errors down the line, or potentially risk some unintentional conversions to `int`.

src/scippnexus/nxdata.py
```diff
@@ -490,8 +494,9 @@ def read_children(self, sel: ScippIndex) -> sc.DataGroup:
         if self._sublogs and ('time' in _to_canonical_select(list(self.sizes), sel)):
             raise sc.DimensionError(
                 "Cannot positionally select time since there are multiple "
-                "time fields. Label-based selection is not supported yet."
+                "time fields."
             )
+
```
SimonHeybrock, line 499: Revert?

src/scippnexus/base.py
jokasimr: But isn't that already handled elsewhere? That should work the same for label based and positional indexing I guess.

SimonHeybrock: I don't think so, because positional indexing works without coords. Give it a try with Scipp vs. ScippNexus (this branch), and see what exception types and error messages you get.

jokasimr: Hm, just noticed the positional indexing behavior is not the same for `scipp.DataGroup` and a scippnexus `File`.
In particular, this does not raise:
```python
with snx.File("60248-2022-02-28_2215.nxs") as f:
    raw = f["entry"][()]
    
raw['some_dimension_name',  :]
```
even if the dimension name does not exist.
But it raises in the example above.

jokasimr:
Okay there are a couple of issues with this implementation.

* I've assumed that the labels are `dimensions` that have coords with the same name. That needs to be fixed.
* It's a bit unclear how the label based slicing should work for hierarchical data, the label based slicing for `DataGroup` in scipp seems to be a bit broken, see example below, so it is unclear how it is supposed to work in scippnexus.

### Something seems wrong with label based slicing for `DataGroup`

Example from the docs:
```python
da = sc.DataArray(
    data=sc.array(dims=['time', 'x'], values=np.random.random((3, 7))),
    coords={
        'x': sc.array(dims=['x'], values=np.linspace(0.1, 0.9, num=7), unit='m'),
        'year': sc.array(dims=['time'], values=[2020, 2023, 2027]),
    },
)
da['year', sc.scalar(2023)].dims # (x,)
```
everything works as expected for `DataArray`.

But for `DataGroup`
```python
dg = sc.DataGroup(
    a=sc.DataArray(
        data=sc.array(dims=['time', 'x'], values=np.random.random((3, 7))),
        coords={
            'x': sc.array(dims=['x'], values=np.linspace(0.1, 0.9, num=7), unit='m'),
            'year': sc.array(dims=['time'], values=[2020, 2023, 2027]),
        },
    ),
)
dg['year', sc.scalar(2023)].dims   # ('time', 'x')
```
the data is not sliced. I'd expect the above to produce a data group with `a` mapped to the sliced `DataArray`.

SimonHeybrock:
Initially `scipp.DataArray` was only support label-based indexing with dimension-coords (i.e., coord name matching dim names). It seems `scipp.DataGroup` still does this... and it might actually be hard/unrealistic to change since we would otherwise need to recursively get a listing of all coords of all children. `scipp.DataGroup` does not "know" about coords, this is not a change I would like to make.

Another approach might be to simply forward all slicing to children, and let them decide if it applies. But that would require not raising on label-based indexing if the coord does not exist, which seems even worse.

jokasimr:
Initially my feeling was that your last suggestion was the most sensible approach, letting each leaf in the data group determine if the slicing is applicable and behave accordingly. This seems to best match most realistic use cases, for example, slicing multiple different detectors / measurements to a particular time interval.

But this has some drawbacks

* no error message if you spell the coord wrong, in scippnexus this behavior would silently load a much larger dataset than intended
* just because cords are named the same does not mean that they refer to the same quantity, they might not even have the same dimension, slicing different leafs in a data group in different dimensions just because their coords are named the same does not seem  convenient

SimonHeybrock:
In conclusion, I think we should not change the `DataGroup` behavior right now (but maybe add a note in the docstring about the difference to `DataArray`).

jokasimr:
Then I guess we should also make the scippnexus label based slicing behave the same, that is, when slicing a group only accept slicing dim-coords, and when slicing a data array accept any coord

SimonHeybrock:
I think that sounds ok for now!

jokasimr:
The tests are failing because I included the test that compares to label slicing a data group in scipp, just to illustrate the problem.
 
 I think this is ready for another review.

jokasimr:
The error messages should probably be changed a bit, they are not very clear.

src/scippnexus/nxdata.py
```diff
@@ -439,6 +444,19 @@ def _add_coords(self, da: sc.DataArray, coords: sc.DataGroup) -> sc.DataArray:
...
+    def _convert_index_to_positional_impl(self, sel):
+        dimcoord, index = sel
+        if dimcoord not in self._children:
+            raise sc.DimensionError(
+                (
+                    f'Invalid slice dimension: \'{dimcoord}\': '
+                    f'no coordinate for that dimension. '
+                    f'Coordinates are {tuple(self._children.keys())}'
+                )
+            )
```
SimonHeybrock, line 456: If we avoid raising here, and just return `sel`, can we avoid the special (and partially incorrect) handling in NXdetector? This would been `sel` is forwarded to the child group (NXevent_data). This would also fix the case of, e.g., and NXdata subgroup in NXdetector or NXmonitor, I think, where the parent does not have the coord, but the subgroup does.

src/scippnexus/nxevent_data.py
```diff
@@ -162,5 +164,20 @@ def assemble(self, children: sc.DataGroup) -> sc.DataArray:
...
+                return ...
+            raise sc.DimensionError(
+                (
+                    f'Invalid slice dimension: \'{dimcoord}\': '
+                    f'no coordinate for that dimension. '
+                    f'Coordinates are {tuple(self._children.keys())}'
+                )
+            )
+        child = self._children[dimcoord][()]
+        return label_based_index_to_positional_index(self.sizes, child, index)
```
SimonHeybrock, line 180: Should this call the same function as the implementation in `NXdata`? No need to repeat, or is there?

jokasimr: You're right, there's no need to repeat it.

SimonHeybrock:
@jokasimr Sorry, I wanted to push to a new branch, but it ended up here! See my suggestion for simplification. Note small difference to exceptions we get from `DataGroup`, but to me the message makes sense, so I think we could take this option?

jokasimr:
No worries!

I don't think it just raises a different exception type, it also raises an exception in cases when data group returns a value. Or am I misreading the errors in the CI test results?

SimonHeybrock:
> No worries!
> 
> I don't think it just raises a different exception type, it also raises an exception in cases when data group returns a value. Or am I misreading the errors in the CI test results?

Which one?

jokasimr:
All of the five test cases hat failed. The failed cases raise on the line 
```python
assert_identical(nx[_slice], dg)
```
If the test reaches that line the scipp data group was sliced without any exception being raised. So if the nexus slicing raises on that line it is not because the error was of the wrong type, but because it raised when slicing the scipp data group didn't.

SimonHeybrock:
> All of the five test cases hat failed. The failed cases raise on the line
> 
> ```python
> assert_identical(nx[_slice], dg)
> ```
> 
> If the test reaches that line the scipp data group was sliced without any exception being raised. So if the nexus slicing raises on that line it is not because the error was of the wrong type, but because it raised when slicing the scipp data group didn't.

I know, but to me the exception looks sensible, e.g., `scipp._scipp.core.DimensionError: 'time' used for indexing not found in dataset dims ('x',).`

jokasimr:
Yeah okay then I misunderstood your original message, I thought you meant that the exception types were different.

Spontaneously I think it's better if slicing in scipp and slicing in scippnexus works the same. Having another set of rules in this case is confusing and should be avoided if possible.

SimonHeybrock:
> Yeah okay then I misunderstood your original message, I thought you meant that the exception types were different.
> 
> Spontaneously I think it's better if slicing in scipp and slicing in scippnexus works the same. Having another set of rules in this case is confusing and should be avoided if possible.

Last week we agreed that consistency within ScippNexus is more relevant than consistency with Scipp. We were considing to later improve Scipp's behavior, if it is inconsistent.

In the above example, what behavior would you expect instead, instead of this exception?

SimonHeybrock (approved):

PR #198, opened 2024-02-15
Support signal given by nested group

### LINKED ISSUE #196: `NXdata` group `@axes` attribute
How `@axes` should be encoded/interpreted has been discussed previously in #145 
   - there, I believe the discussion was about `@axes` applied to the `dataset` named 'DATA'

Comparing what the NeXus standard says regarding how the `@axes` attribute should be encoded when applied to the 'DATA' `dataset` and its containing `NXdata` group, there is some ambiguity.

Namely, notice that both use the term 'array'
> Defines the names of the coordinates (independent axes) for this data set as a colon-delimited array.
>
> &mdash; [`NXdata/data @axes`](https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-data-axes-attribute)

> When axes contains multiple strings, it must be saved as an actual array of strings and not a single comma separated string.
>
>  &mdash; [`NXdata @axes`](https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-axes-attribute) 

Where this becomes ambiguous, is in that HDF5 allows for array-valued attributes (which [h5py reads as numpy arrays](https://docs.h5py.org/en/stable/high/attr.html)), and `scippnexus` has interpreted both 'colon-delimited array' and 'actual array of strings' to specify a single string with axes names separated by literal ':' characters.

## Issue
If someone interprets 'actual array of strings' to mean 'array-valued attribute with string elements', then creates a NeXus HDF5 file, like the one attached; that file can not be opened by `scippnexus` due to the use of the `split` method here:
https://github.com/scipp/scippnexus/blob/58ffbbe67e90bf3341a7589e89e00e36859ba712/src/scippnexus/nxdata.py#L160

## Possible solution
Add a check that `self._signal_axes` is a `str` before attempting to use `.split`, or alternately a check if it is a numpy array, e.g., replacing the quoted line 160 by
```python
            self._signal_axes = tuple(self._signal_axes if isinstance(self._signal_axes, np.ndarray) else self._signal_axes.split(':'))
```

## Alternate solution
If the interpretation of 'actual array of strings' should be limited to 'colon-delimited array' like the 'DATA' dataset attribute, 
petition the NeXus committee to clarify their boxed-note for the group attribute.

SimonHeybrock:
NeXus probably allows using NXdata (or related, such as NXlog) in place of "signal" datasets. This is a small fix to the NXdata implementation arounding handling shape/dims definition making this possible.

Fixes #196, I think.


g5t:
This change does indeed appear to fix my issue in #196
Thanks!

image

PR #205, opened 2024-04-23
[Fix] issue 204 by removing trailing forward slashes in string selectors

### LINKED ISSUE #204: A trailing `'/'` in a file selector path raises a `KeyError`
Accessing a file via `scippnexus` with a relative or absolute path that ends in an extraneous `'/'` raises a `KeyError` due to attempting to access the empty `path` `''` in the following code snippet 

https://github.com/scipp/scippnexus/blob/86a4bfe37e4cc1d1cdf574176c0986ea345f5785/src/scippnexus/base.py#L382-L387

I expected the trailing forward slash to not raise an error; and for accessing, e.g., `'/entry'` and `/entry/'`, to result in the same object.

Is there a reason that a trailing `'/'` _should_ raise an error?

g5t:
One possible fix for #204 if any number of trailing forward slashes _should not_ raise an error.

Fixes #204.


SimonHeybrock (commented):
Thanks, look reasonable. Do you feel like adding a unit test? If not, we can take care of that.

MridulS:
The doc failure should be fixed by https://github.com/scipp/scippnexus/pull/207

g5t:
I've added a unit test which will fail with a `KeyError` before my proposed fix.
It may be more-restrictive than necessary, since it only passes the assert if **the same** object is returned with and without a trailing forward slash in its path.

SimonHeybrock (approved):
Thanks!

Issue #195, opened 2024-02-12
Deprecation warning with Python 3.12 


YooSunYoung:
I wanted to test python 3.12 for other packages,
then I found out ``scippnexus`` was using ``dateutil.parser`` which throws DeprecationWarning in python 3.12.

But in the copier answer says it supports up to python 3.12
so I was wondering if it was tested with python 3.12 and I should just pin some dependencies...

SimonHeybrock:
To be clear, are you saying that ScippNexus is not working, or is it simply causing a deprecation warning?

YooSunYoung:
> To be clear, are you saying that ScippNexus is not working, or is it simply causing a deprecation warning?

It works, but it causes deprecation warning...! I'll update the title to be more clear.

MridulS:
I think this is fixed upstream https://github.com/dateutil/dateutil/issues/1314

MridulS:
I think we can use https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat directly? One less dependency?

PR #206, opened 2024-04-23
 Use stdlib datetime to parse instead of dateutil 

MridulS:
One way for fixing https://github.com/scipp/scippnexus/issues/195.

There are indeed some issues with using the stdlib version in <py3.11 https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat but the tests seems to pass (not sure how comprehensive they are) and hopefully we should move to py3.11 soon enough.


SimonHeybrock:
Need to drop Python 3.9 first?

MridulS:
Urghh yeah, this actually works only py3.11+ so would need to wait some time.

Issue #208, opened 2024-04-24
Repeated forward slashes become absolute paths


g5t:
Following up on #204 and the fix in #205, I noticed that a path with repeated forward slashes will silently ignore any valid path before the last such `'/'` and treat the rest of the path specification as an absolute path. E.g., a NeXus HDF5 file with an `NXinstrument` group at `/entry/instrument` opened as `file` will pass the following assert statement

```python
assert id(file['/entry/instrument']) == id(file['/entry//entry/instrument'])
```

this behavior is contrary to file system paths under [unix-like systems which treat repeated path separators as if they were singular](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_271).

MridulS:
Maybe we can use `pathlib` to santize these things? I know they aren't path objects, but it should take care of all these corner cases.

jl-wynen:
Yes:
```python
PurePosixPath('/entry//entry/instrument').as_posix()
```
Also takes care of the final slash

PR #209, opened 2024-04-24
MAINT: use pathlib for path manipulation

MridulS:
This should fix https://github.com/scipp/scippnexus/issues/208 but I feel like there should be a better solution.

### DIFF
--- a/src/scippnexus/base.py
+++ b/src/scippnexus/base.py
@@ -7,6 +7,7 @@ import inspect
 import warnings
 from collections.abc import Mapping
 from functools import lru_cache
+from pathlib import PurePosixPath
 from types import MappingProxyType
 from typing import Any, Dict, Iterator, List, Optional, Tuple, Union, overload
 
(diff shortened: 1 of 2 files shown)


src/scippnexus/base.py
```diff
@@ -380,13 +381,15 @@ def __getitem__(self, sel):
             # create the parent group, to ensure that fields get the correct properties
             # such as sizes and dtype.
             if '/' in sel:
-                if sel.startswith('/'):
-                    return self.file[sel[1:]]
-                elif sel.endswith('/'):
-                    return self[sel[:-1]]
+                sel_path = Path(sel)
```
jl-wynen, line 384: This depends on the OS. You need to use `PosixPath` here. Or better `AbstractPosixPath` because it doesn't refer to a real file.

This is one reason why only testing on Linux is not enough... But that is a different discussion.

jl-wynen (changes requested):

MridulS:
There are unrelated windows error in scippnexus btw.
One of the errors in the CI run with windows, let's use https://github.com/scipp/copier_template/pull/166 for the windows discussion.

```
E           RuntimeError: Coordinate must be monotonically increasing or decreasing for label-based indexing.
```

PR #210, opened 2024-04-29
Update python version in nightly.txt requirements after py3.9 was dropped

nvaytet:
It seems that in #207 , `requirements/nightly.in` was updated (use python 3.10 instead of 3.9), but `requirements/nightly.txt` was not updated accordingly.

We re-run `tox -e deps` to fix this here.


requirements/nightly.txt
```diff
@@ -15,9 +15,9 @@ numpy==1.26.4
     #   scipy
 python-dateutil==2.9.0.post0
     # via -r nightly.in
-scipp @ https://github.com/scipp/scipp/releases/download/nightly/scipp-nightly-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
+scipp @ https://github.com/scipp/scipp/releases/download/nightly/scipp-nightly-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```
nvaytet, line 18: This is the important change

MridulS: Yeah we should try to lock this somehow https://github.com/scipp/copier_template/issues/163 as it is currently senstive to the local env used to run tox.

YooSunYoung (approved):

Issue #222, opened 2024-08-13
Docstring of `Group` methods/properties (and others?) don't show in docs


SimonHeybrock:
Compare https://scipp.github.io/scippnexus/generated/classes/scippnexus.Group.html to `Group` in https://github.com/scipp/scippnexus/blob/main/src/scippnexus/base.py. Why is no member documentation shown for this class?

jokasimr:
Older scippnexus releases have the same problem, checked back to version 23.11, before that we didn't store the docs build in the release artifacts. So it seems the issue is not because of some recent change.

PR #223, opened 2024-08-13
Fix typo in autodoc generation, point towards right template

MridulS:
Fixes https://github.com/scipp/scippnexus/issues/222

### DIFF
--- a/docs/api-reference/index.md
+++ b/docs/api-reference/index.md
@@ -95,7 +95,7 @@
 
 .. autosummary::
    :toctree: ../generated/classes
-   :template: scipp-class-template.rst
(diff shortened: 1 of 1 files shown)


jokasimr:
There's another reference to `scipp-class-template.rst` in the same file: https://github.com/scipp/scippnexus/blob/5fd2af17e4d94b369b1a2d5b0695f1f11ecfc5b9/docs/api-reference/index.md?plain=1#L98

I guess we should remove that one also?

MridulS:
> I guess we should remove that one also?

Yeah, thanks for the catch!

jokasimr (approved):

PR #229, opened 2024-09-04
Fix NXcite subgroup tripping up NXdetector loading

SimonHeybrock:
This happens with some CODA files (seen for NMX).

Also remove a global warning ignore.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,6 @@ addopts = """
 testpaths = "tests"
 filterwarnings = [
   "error",
-  "ignore:Failed to load :UserWarning",
 ]
 markers = [
     "externalfile",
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -61,7 +61,6 @@ addopts = """
 testpaths = "tests"
 filterwarnings = [
   "error",
-  "ignore:Failed to load :UserWarning",
```
jokasimr, line 64: I wonder what changed so that we don't see this anymore.

jokasimr (approved):

pyproject.toml
SimonHeybrock: See second commit.

PR #238, opened 2024-09-09
Allow passing a h5py file when opening a file

### LINKED ISSUE #236: Support passing an already open h5py file
This should be possible:
```python
with h5py.File(...) as f:
    with snx.File(f, ...) as nx:
        ...
```

But currently, `snx.File.__init__` simply forwards all arguments to `h5py.File` which does not support this usage pattern.

jl-wynen:
Fixes #236

### DIFF
--- a/src/scippnexus/file.py
+++ b/src/scippnexus/file.py
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
-# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
 # @author Simon Heybrock
+import io
+import os
 from contextlib import AbstractContextManager
 
 import h5py
@@ -17,6 +19,7 @@ from .typing import Definitions
 class File(AbstractContextManager, Group):
     def __init__(
         self,
+        name: str | os.PathLike[str] | io.BytesIO | h5py.Group,
         *args,
         definitions: Definitions | DefaultDefinitionsType = DefaultDefinitions,
         **kwargs,
@@ -27,6 +30,9 @@ class File(AbstractContextManager, Group):
 
         Parameters
         ----------
+        name:
+            Path, Bytes object, or h5py.Group.
+            Specifies the file to open.
         definitions:
             Mapping of NX_class names to application-specific definitions.
             The default is to use the base definitions as defined in the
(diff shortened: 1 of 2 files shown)


src/scippnexus/file.py
```diff
@@ -27,22 +30,35 @@ def __init__(
 
         Parameters
         ----------
+        name:
+            Path, Bytes object, or h5py.Group.
```
nvaytet, line 34: Can we leave this line out if it's already specified in the type hints? Would avoid it accidentally becoming out of date if we change something in the future...

src/scippnexus/file.py
```diff
@@ -17,6 +19,7 @@
 class File(AbstractContextManager, Group):
     def __init__(
         self,
+        name: str | os.PathLike[str] | io.BytesIO | h5py.Group,
```
nvaytet, line 22: If the file is open, is it a `h5py.Group` or a `h5py.File`?

tests/file_test.py
```diff
@@ -0,0 +1,65 @@
...
+def is_closed(file: snx.File) -> bool:
+    try:
+        file.create_class('x', 'NXentry')
+    except ValueError as err:
+        return 'Unable to synchronously create group' in err.args[0]
+    return False
+
+
+@pytest.mark.parametrize('path_type', [str, Path])
+def test_load_entry_from_filename(tmp_path, path_type):
```
nvaytet, line 22: `tmp_path`: I hate `pytest` and all its hidden conventions/pre-defined objects. It took me a while to figure out where `tmp_path` was defined... :-(

tests/file_test.py
```diff
@@ -0,0 +1,65 @@
...
+    with h5.File(buffer, 'w') as f:
+        snx.create_class(f, 'entry', snx.NXentry)
+    buffer.seek(0)
+
+    with snx.File(buffer, 'r') as f:
+        assert f.keys() == {'entry'}
+    assert is_closed(f)
+
+
+def test_load_entry_from_h5py_group_toor(tmp_path):
```
nvaytet, line 42:
```suggestion
def test_load_entry_from_h5py_group_root(tmp_path):
```

tests/file_test.py
```diff
@@ -0,0 +1,65 @@
...
+    with h5.File('test.nxs', 'w', driver='core', backing_store=False) as h5_file:
```
nvaytet, line 43: I can't quite figure out if this is creating a file on disk or not. From the `h5py` docs for `backing_store`, it says "If False, any changes are discarded when the file is closed."
Does this mean that if the file did not exist, and you did not make changes to it, then no file will be created, or will an empty file be created?

tests/file_test.py
```diff
@@ -0,0 +1,65 @@
...
+        with snx.File(h5_file['entry']) as snx_file:
+            assert snx_file.keys() == {'user'}
+        # Not great, but we don't want to close the h5py file ourselves.
+        assert not is_closed(snx_file)
+
+
+def test_file_from_h5py_group_does_not_allow_extra_args(tmp_path):
+    with h5.File('test.nxs', 'w', driver='core', backing_store=False) as h5_file:
+        snx.create_class(h5_file, 'entry', snx.NXentry)
+        with pytest.raises(TypeError):
```
nvaytet, line 64:
```suggestion
        with pytest.raises(TypeError, match='Cannot provide both h5py.File and other arguments'):
```
?

src/scippnexus/file.py
jl-wynen: `h5py.File`. But that is a subclass of `h5py.group`. So this annotation is enough and also supports passing in a subgroup.

tests/file_test.py
jl-wynen: I don't actually know. This is used everywhere else in the tests. I just copied it from there.

nvaytet (approved):

PR #242, opened 2024-09-26
fix: warn if missing vector in transformation

jokasimr:
Fixes https://github.com/scipp/essreflectometry/issues/81

* Is `NexusStructureError` the right kind of error to raise if the transform is missing the `vector` attribute?


src/scippnexus/base.py
```diff
@@ -418,7 +418,12 @@ def isclass(x):
             if (resolved := maybe_resolve(self['depends_on'], depends_on)) is not None:
                 dg['resolved_depends_on'] = resolved
 
-        return maybe_transformation(self, value=dg, sel=sel)
+        try:
+            dg = maybe_transformation(self, value=dg, sel=sel)
+        except (sc.DimensionError, NexusStructureError) as e:
+            self._warn_fallback(e)
```
SimonHeybrock, line 424: I think this is actually the wrong warning, since it will say "failed to load as NXlog", but that part worked. The subsequent parsing of the log as a transform failed. Maybe moving the try/catch into `maybe_transformation` is better after all?

src/scippnexus/nxtransformations.py
```diff
@@ -87,6 +87,8 @@ def make_transformation(
                 value = value['value']
             if isinstance(value, sc.DataGroup):
                 return value
+            if self.attrs.get('vector') is None:
+                raise NexusStructureError('A transformation needs a vector attribute')
```
SimonHeybrock, line 91: Yes, I think using this error is ok.

jokasimr: How about using `TransformationError`?

src/scippnexus/nxtransformations.py
```diff
@@ -117,7 +119,12 @@ def make_transformation(
                     depends_on_to_relative_path(depends_on, self._obj.parent.name)
                 )
             return transform
-        except (sc.DimensionError, sc.UnitError, TransformationError):
+        except (sc.DimensionError, sc.UnitError, TransformationError) as e:
+            msg = (
+                f"Failed to load {self.name} as {NXtransformations.__name__}: {e} "
+                "Falling back to returning non transformed value."
```
SimonHeybrock, line 125: `NXtransformations` is the parent group and loading that did not fail.

```suggestion
                f"Failed to convert {self.name} into a transformation: {e} "
                "Falling back to returning non underlying value."
```

src/scippnexus/nxtransformations.py
```diff
@@ -87,6 +87,8 @@ def make_transformation(
                 value = value['value']
             if isinstance(value, sc.DataGroup):
                 return value
+            if self.attrs.get('vector') is None:
+                raise TransformationError('A transformation needs a vector attribute.')
```
SimonHeybrock, line 91: Better put this into `self.vector`?

src/scippnexus/base.py
jokasimr: Moved the test into `maybe_transformation` instead, and made it emit the warning if the transform loading errored out for any reason. Is that desired?

SimonHeybrock (approved):

PR #244, opened 2024-10-15
fix: nxlog use time field as signal if missing

### LINKED ISSUE #230: Support NXlog without value dataset
This appears to be a valid case, used for top-dead-center chopper timestamps. Currently there is a warning triggered. Probably all there is to do is avoid the warning in this case, by using the `time` dataset as the `signal`?

Should this return the `time` as a `DataArray`, or simply as a `Variable` (as it is now)?

jokasimr:
Fixes #230

### DIFF
--- a/src/scippnexus/nxdata.py
+++ b/src/scippnexus/nxdata.py
@@ -486,7 +486,7 @@ class NXlog(NXdata):
             attrs=attrs,
             children=children,
             fallback_dims=('time',),
-            fallback_signal_name='value',
+            fallback_signal_name='value' if 'value' in children else 'time',
         )
 
     def read_children(self, sel: ScippIndex) -> sc.DataGroup:
--- a/tests/nxlog_test.py
+++ b/tests/nxlog_test.py
@@ -52,16 +52,14 @@ def test_nxobject_log(h5root):
     assert sc.identical(log[...], da)
 
 
-@pytest.mark.filterwarnings("ignore:Failed to load /entry/log:UserWarning")
-def test_nxlog_with_missing_value_triggers_fallback(nxroot):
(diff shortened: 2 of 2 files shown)


tests/nxlog_test.py
```diff
@@ -52,16 +52,14 @@ def test_nxobject_log(h5root):
     assert sc.identical(log[...], da)
 
 
-@pytest.mark.filterwarnings("ignore:Failed to load /entry/log:UserWarning")
 def test_nxlog_with_missing_value_triggers_fallback(nxroot):
```
SimonHeybrock, line 55: Test name needs update

src/scippnexus/nxdata.py
```diff
@@ -489,6 +489,13 @@ def __init__(self, attrs: dict[str, Any], children: dict[str, Field | Group]):
             fallback_signal_name='value',
         )
 
+    def _init_signal(self, *args, children, **kwargs):
+        super()._init_signal(*args, children=children, **kwargs)
+        if self._signal is None:
+            if 'time' in children:
+                self._signal_name = 'time'
+                self._signal = children['time']
```
SimonHeybrock, line 497: Could this simply use
```python
fallback_signal_name='value' if 'value' in children else 'time'`
```
?

jokasimr: That's better :+1:

jokasimr: But, there's no argument to `_init_signal` called `fallback_signal_name`

SimonHeybrock: My point was to change `__init__` and remove `_init_signal`.

jokasimr: Aha I see

SimonHeybrock (approved):

PR #248, opened 2024-11-09
Add depends_on path to error output

ggoneiESS:
It's hard to diagnose current web output as there is no simple way to see what is being warned about

### DIFF
--- a/src/scippnexus/nxtransformations.py
+++ b/src/scippnexus/nxtransformations.py
@@ -280,7 +280,8 @@ def parse_depends_on_chain(
             chain.transformations[transform.name] = transform[()]
     except KeyError as e:
         warnings.warn(
-            UserWarning(f'depends_on chain {depends_on} references missing node {e}'), stacklevel=2
+            UserWarning(f'depends_on chain {depends_on} references missing node {e}'),
+            stacklevel=2,
         )
         return None
     return chain
--- a/src/scippnexus/nxtransformations.py
+++ b/src/scippnexus/nxtransformations.py
@@ -280,7 +280,7 @@ def parse_depends_on_chain(
             chain.transformations[transform.name] = transform[()]
     except KeyError as e:
         warnings.warn(
-            UserWarning(f'depends_on chain references missing node {e}'), stacklevel=2
+            UserWarning(f'depends_on chain {depends_on} references missing node {e}'), stacklevel=2
         )
         return None
     return chain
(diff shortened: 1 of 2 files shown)


SimonHeybrock (approved):
Thanks!

src/scippnexus/nxtransformations.py
```diff
@@ -280,7 +280,8 @@ def parse_depends_on_chain(
             chain.transformations[transform.name] = transform[()]
     except KeyError as e:
         warnings.warn(
-            UserWarning(f'depends_on chain references missing node {e}'), stacklevel=2
+            UserWarning(f'depends_on chain {depends_on} references missing node {e}'),
```
nvaytet, line 283: No sure what this adds? I think the depends_on was already in the warning message as `e`?

Old error message: `UserWarning: depends_on chain references missing node 'transform'`
New error message: `UserWarning: depends_on chain transform references missing node 'transform'`

The `KeyError` is just the name of the missing entry.

ggoneiESS: I think that's a lucky example, what I'm interested in is which node is causing the problem, and I get a message like so:

`WARNING  root:fileloading_test.py:31 UserWarning: depends_on chain /entry/instrument/detector_panel_2/transformations/stageZ references missing node 'depends_on'`

Run against `443503_00032754.hdf`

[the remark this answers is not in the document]
ggoneiESS: I ran it locally like so `UserWarning(f'depends_on chain references missing node {e}  {depends_on}')`, and that was good enough for me then although it doesn't read as nicely.  If that makes it easier to do the matching though you could do that, in case you match multiple warnings using wildcards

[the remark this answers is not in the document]
ggoneiESS: done, let's see if it works

tests/nxtransformations_test.py
```diff
@@ -729,9 +729,9 @@ def test_compute_positions_warns_if_depends_on_is_dead_link(h5root):
     detector = create_detector(instrument)
     snx.create_field(detector, 'depends_on', sc.scalar('transform'))
     root = make_group(h5root)
-    with pytest.warns(UserWarning, match='depends_on chain references missing node'):
+    with pytest.warns(UserWarning, match='depends_on chain .* references missing node'):
```
MridulS, line 732:
```suggestion
    warning_regex = r"depends_on chain .* missing node"
    with pytest.warns(UserWarning, match=warning_regex):
```

tests/nxtransformations_test.py
```diff
@@ -729,9 +729,9 @@ def test_compute_positions_warns_if_depends_on_is_dead_link(h5root):
...
         loaded = root[()]
-    with pytest.warns(UserWarning, match='depends_on chain references missing node'):
+    with pytest.warns(UserWarning, match='depends_on chain .* references missing node'):
```
MridulS, line 734:
```suggestion
    with pytest.warns(UserWarning, match=warning_regex):
```

PR #253, opened 2024-11-19
Optimise depends_on chain resolution by using h5py

jl-wynen:
This reduces the time it takes to resolve `depends_on` chains by locating the target using h5py and a custom path resolver. It avoids creating intermediate `snx.Group` objects which is expensive.

For loading the 45 BIFROST analysers using 
```python
def load(fname):
    start = time.perf_counter()
    with snx.File(fname) as f:
        instrument = f['entry']['instrument']
        groups = instrument[snx.NXcrystal]
        start_load = time.perf_counter()
        crystals = {k: v[()] for k, v in groups.items()}
        end_load = time.perf_counter()
    closed = time.perf_counter()
    print(f'total: {closed-start:.3f}s   load: {end_load-start_load:.3f}s')
```
I get these results on my machine:
```
# Baseline:       total: 1.205s   load: 1.082s
# Custom resolve: total: 0.345s   load: 0.306s
```

### DIFF
--- a/src/scippnexus/field.py
+++ b/src/scippnexus/field.py
@@ -81,6 +81,31 @@ def _as_datetime(obj: Any):
     return None
 
 
+_scipp_dtype = {
+    np.dtype('int8'): sc.DType.int32,
+    np.dtype('int16'): sc.DType.int32,
+    np.dtype('uint8'): sc.DType.int32,
+    np.dtype('uint16'): sc.DType.int32,
+    np.dtype('uint32'): sc.DType.int32,
+    np.dtype('uint64'): sc.DType.int64,
+    np.dtype('int32'): sc.DType.int32,
+    np.dtype('int64'): sc.DType.int64,
+    np.dtype('float32'): sc.DType.float32,
+    np.dtype('float64'): sc.DType.float64,
+    np.dtype('bool'): sc.DType.bool,
+}
+
+
+def _dtype_fromdataset(dataset: H5Dataset) -> sc.DType:
+    return _scipp_dtype.get(dataset.dtype, sc.DType.string)
+
+
+def _squeezed_field_sizes(dataset: H5Dataset) -> dict[str, int]:
+    if (shape := dataset.shape) == (1,):
+        return {}
+    return {f'dim_{i}': size for i, size in enumerate(shape)}
+
+
 @dataclass
 class Field:
     """NeXus field.
@@ -92,6 +117,12 @@ class Field:
     sizes: dict[str, int] | None = None
     dtype: sc.DType | None = None
     errors: H5Dataset | None = None
+
+    def __post_init__(self) -> None:
+        if self.sizes is None:
+            self.sizes = _squeezed_field_sizes(self.dataset)
+        if self.dtype is None:
+            self.dtype = _dtype_fromdataset(self.dataset)
 
     @cached_property
     def attrs(self) -> dict[str, Any]:
--- a/src/scippnexus/nxtransformations.py
+++ b/src/scippnexus/nxtransformations.py
@@ -36,10 +36,12 @@ such as streamed NXlog values received from a data acquisition system.
 from __future__ import annotations
 
 import warnings
+from collections.abc import Mapping
 from dataclasses import dataclass, field, replace
 from pathlib import PurePosixPath
 from typing import Literal
 
+import h5py
 import numpy as np
 import scipp as sc
 from scipp.scipy import interpolate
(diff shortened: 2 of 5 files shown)


src/scippnexus/nxtransformations.py
```diff
@@ -274,7 +330,7 @@ def parse_depends_on_chain(
     depends_on = depends_on.value
     try:
         while depends_on != '.':
-            transform = parent[depends_on]
+            transform = _locate_depends_on_target(parent, depends_on)
             parent = transform.parent
```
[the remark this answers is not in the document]
jl-wynen: I did this now. But it doesn't affect performance.

src/scippnexus/field.py
```diff
@@ -93,6 +118,12 @@ class Field:
     dtype: sc.DType | None = None
     errors: H5Dataset | None = None
 
+    def __post_init__(self) -> None:
+        if self.sizes is None:
+            self.sizes = _squeezed_field_sizes(self.dataset)
+        if self.dtype is None:
+            self.dtype = _dtype_fromdataset(self.dataset)
+
```
jl-wynen, line 126: Please check that this is sound!

SimonHeybrock: I feared that there might be some code that relies on `field.sizes` being `None` (e.g., to do custom init logic only if dims are not set yet) before calling `NXobject.__init__`, but I could not find any (looking mainly at `nxdata.py`). So I think it should be ok?

src/scippnexus/nxtransformations.py
```diff
@@ -266,16 +275,67 @@ def compute(self) -> sc.Variable | sc.DataArray:
...
+        return path
+
+    base_parts = list(cwd.parts)
+    path_parts = [segment for segment in path.parts if segment != '.']
+    while path_parts[0] == '..':
+        if not path_parts:
+            raise ValueError(f"Relative path beyond root: '{path}'")
+        base_parts.pop(-1)
+        path_parts.pop(0)
+    return PurePosixPath(*base_parts, *path_parts)
```
SimonHeybrock, line 295: Can we use something like https://github.com/scipp/scippnexus/blob/349c8df43df0771f930bec12b13db7b912052fc3/src/scippnexus/field.py#L39-L42?

src/scippnexus/nxtransformations.py
```diff
@@ -266,16 +275,67 @@ def compute(self) -> sc.Variable | sc.DataArray:
...
+    file: h5py.File,
+    base_path: PurePosixPath,
+    depends_on: PurePosixPath,
+    definitions: Mapping[str, type] | None,
+) -> tuple[Field | Group, PurePosixPath]:
+    """Find the target of a depends_on link.
+
+    The returned object is equivalent to calling ``parent[depends_on]``
+    in the context of transformations.
+    This function does not work in general because it does not handle NXdata attributes.
```
SimonHeybrock, line 308: Not sure why you mention NXdata attributes? What are you referring to?

src/scippnexus/nxtransformations.py
```diff
@@ -266,16 +275,67 @@ def compute(self) -> sc.Variable | sc.DataArray:
...
+    file = parent.underlying.file
+    base_path = PurePosixPath(parent.underlying.name)
     try:
         while depends_on != '.':
-            transform = parent[depends_on]
-            parent = transform.parent
+            transform, base_path = _locate_depends_on_target(
+                file, base_path, PurePosixPath(depends_on), parent.definitions
+            )
             depends_on = transform.attrs['depends_on']
```
SimonHeybrock, line 339: As we are loading the transform anyway, and the loaded `Transform` has access to its parent via its `depends_on`, do we need the `base_path` returned from `_locate_depends_on_target`? It seems like a duplicate control flow, unless I am missing something.

src/scippnexus/nxtransformations.py
jl-wynen: What is `posixpath`?

src/scippnexus/nxtransformations.py
jl-wynen: I thought this is what the comment here https://github.com/scipp/scippnexus/blob/349c8df43df0771f930bec12b13db7b912052fc3/src/scippnexus/base.py#L371-L373 refers to.

src/scippnexus/nxtransformations.py
jl-wynen: I did this to avoid accessing the parent because of https://github.com/scipp/scippnexus/pull/253#discussion_r1848149190

src/scippnexus/nxtransformations.py
SimonHeybrock: `import posixpath`

src/scippnexus/nxtransformations.py
SimonHeybrock: Yes, but NXdata is just one case (it is a baseclass for some but not all other NX* implementations).

src/scippnexus/nxtransformations.py
SimonHeybrock: My point was: We are loading the transform anyway (in the line below), so the parent path is just a Python string stored in a Python class? There should be no more `snx.Group` stuff at the level.

jl-wynen: I see. The `DependsOn` object could even make the whole implementation simpler.
But it doesn't actually work because `transform[()]` can return `DataArray` if the transform is broken in the file. In that case, the current code still builds a complete transformation chain that contains data arrays instead of `Transform` objects. But those don't carry a `depends_on` attribute. See, e.g., `test_compute_transformation_warns_if_transformation_missing_vector_attr`.

So I don't see an alternative to tracking the parent path manually here that doesn't change the behaviour.

SimonHeybrock (approved):

PR #254, opened 2024-11-25
Detect cycles in depends_on

jl-wynen:
We have integration tests with broken files that have circular `depends_on` graphs. ScippNeXus currently runs in an infinite loop in those cases.

This PR adds
- cycles in `depends_on` chains are detected and lead to `ValueError`s.
- Self-referencing `depends_on` are treated as `depends_on == '.'`. While this does not comply with the standard, I see no reason not to do this.

### DIFF
--- a/tests/nxtransformations_test.py
+++ b/tests/nxtransformations_test.py
@@ -931,3 +931,34 @@ def test_compute_positions_terminates_with_depends_on_to_self_relative(h5root):
             unit='m',
         ),
     )
+
+
+def test_compute_positions_raises_for_depends_on_cycle(h5root):
+    instrument = snx.create_class(h5root, 'instrument', snx.NXinstrument)
+    detector = create_detector(instrument)
+    snx.create_field(detector, 'x_pixel_offset', sc.linspace('xx', -1, 1, 2, unit='m'))
+    snx.create_field(detector, 'y_pixel_offset', sc.linspace('yy', -1, 1, 2, unit='m'))
+    detector.attrs['axes'] = ['xx', 'yy']
+    detector.attrs['x_pixel_offset_indices'] = [0]
+    detector.attrs['y_pixel_offset_indices'] = [1]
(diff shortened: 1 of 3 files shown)


tests/nxtransformations_test.py
```diff
@@ -829,3 +829,136 @@ def test_compute_transformation_warns_if_transformation_missing_vector_attr(
...
+    value1.attrs['offset_units'] = str(offset.unit)
+    value1.attrs['vector'] = vector.value
+    value2 = snx.create_field(transformations, 't2', value.to(unit='cm'))
+    value2.attrs['depends_on'] = 't1'
+    value2.attrs['transformation_type'] = 'translation'
+    value2.attrs['vector'] = vector.value
+
+    root = make_group(h5root)
+    with pytest.raises(ValueError, match="Circular depends_on"):
+        _ = root[()]
```
nvaytet, line 964: Thanks.

Did you try loading a NMX file from coda to check that the error is raised?

jl-wynen: I did now and it does find a cycle:
```
['/entry/instrument/detector_panel_0/transformations/orientation', '/entry/instrument/detector_panel_0/transformations/axis6', '/entry/instrument/detector_panel_0/transformations/axis5', '/entry/instrument/detector_panel_0/transformations/axis4', '/entry/instrument/detector_panel_0/transformations/axis3', '/entry/instrument/detector_panel_0/transformations/axis2', '/entry/instrument/detector_panel_0/transformations/axis1', '/entry/instrument/detector_panel_0/transformations/stageZ']
```

nvaytet (approved):

SimonHeybrock:
> I don't think that typos are a big concern. How many files are written by hand?

I guess close to 100% of the code or config files used to write files is written by hand. I am not sure what you are saying. I think self-loops are just as bad as other loops?

PR #255, opened 2024-11-29
Do not allow self-ref in depends_on

jl-wynen:
See https://github.com/scipp/scippnexus/pull/254#issuecomment-2505934364

@SimonHeybrock happy?


SimonHeybrock (approved):
Thanks!

PR #261, opened 2025-01-30
Add quiet flag for warnings

### LINKED ISSUE #260: Add a `quiet` argument on `load` to suppress warnings?
When opening non-ESS files, the list of warnings from loading a file can be very long.
Can we add a `quiet` or `suppress_warnings` argument to the `load` function (which would be `False` by default)?

nvaytet:
Using `snx.load(filename, quiet=True)` suppressed all warnings raised when loading files.

Fixes #260

### DIFF
--- a/src/scippnexus/_load.py
+++ b/src/scippnexus/_load.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 import contextlib
 import io
+import warnings
 from os import PathLike
 
 import h5py as h5
@@ -23,6 +24,7 @@ def load(
     root: str | None = None,
     select: ScippIndex = (),
     definitions: Definitions | DefaultDefinitionsType = DefaultDefinitions,
+    quiet: bool = False,
 ) -> sc.DataGroup | sc.DataArray | sc.Dataset:
     """Load a NeXus file.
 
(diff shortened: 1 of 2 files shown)


src/scippnexus/_load.py
```diff
@@ -83,16 +85,20 @@ def load(
...
     -------
     :
         The loaded data.
     """
-    with _open(filename, definitions=definitions) as group:
-        if root is not None:
-            group = group[root]
-        return group[select]
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore" if quiet else "default")
```
jl-wynen, line 97: I think this resets user-defined filters. You need to not call `simplefilter` if `not quiet`. But I may be wrong. Did you check?

src/scippnexus/_load.py
jl-wynen, line 1: There are some issues with this approach:

- It affects all warnings, even those that may not be related to file structure but, e.g., arise from dependencies.
- It works by changing a global variable. So if this code is used in a multi-threaded context, e.g., in Sciline with Dask, the filter also affects warnings that don't arise from ScippNexus.

Those issues can be somewhat mitigated by filtering on the module where the warning is raised. But you can still get race conditions with multi-threading.

If you want to go with this approach, please add a clear docstring explaining the caveats.

SimonHeybrock: Would it make sense to consider using logging instead of warnings?

jl-wynen: Not if you want to treat warnings as errors in some contexts.

SimonHeybrock: Have we considered putting the filtering code into userland, i.e., not change ScippNexus?

nvaytet: It's an option, I just thought having a flag would be the most user friendly mechanism.

PR #264, opened 2025-02-04
Add more verbose message to attribute error

ggoneiESS:
It is often not enough to know that there is an error for specific instances - the path that is at issue also needs to be reported.  

This is similar to https://github.com/scipp/scippnexus/pull/248 and so now most cases are covered it seems.


SimonHeybrock (commented):
Thanks! But could I ask you undo the reformat that changes all the quotes?

ggoneiESS:
yes, you may be able to see from the commits Ruff has kept to its name and not been smooth...should pass now and I've also added some additional paths for `depends_on` warnings

PR #276, opened 2025-07-28
Lower bounds

jokasimr:


pyproject.toml
```diff
@@ -31,17 +31,17 @@ requires-python = ">=3.10"
 # Run 'tox -e deps' after making changes here. This will update requirement files.
 # Make sure to list one dependency per line.
 dependencies = [
-    "python-dateutil",
+    "python-dateutil>=2.8",
     "scipp>=24.02.0",
-    "scipy",
-    "h5py",
+    "scipy>=1.10.0",
+    "h5py>=3.12",
```
SimonHeybrock, line 37: Can you check if Mantid works with this? If so, remove https://github.com/scipp/scippnexus/blob/60d60237c00c3427386712099f83e165f29c1dfb/src/scippnexus/field.py#L221-L229 else this needs to be reverted, I think?

MridulS: I think this was fixed in mantid in https://github.com/mantidproject/mantid/pull/34806 ?

PR #298, opened 2026-02-10
feat: visualize transform chains

jokasimr:
Fixes #292


jokasimr:
Screenshot from 2026-02-10 12-52-17

nvaytet:
I think it would be useful to show the values as well? How many meters it was translated by in what direction?
Also degrees of rotation? (for rotations, maybe if we can see it's around a specific axis we can show that; otherwise, I am not sure showing the rotation matrix is so useful?)

Finally, I wonder if placing things at scale in the graph would be useful/possible for the translations? You could then build some sort of instrument layout with relative spacings between components?

jokasimr:
I added fields for the transformation vector and the magnitude.

Placing things at scale I think wouldn't be so useful here. I interpreted this to be a more abstract illustration to debug/understand transformations in a nexus file. It would be difficult to make it look good, because some translations might be tiny and other might be large, and it would only cover the translations anyway (since we can't rotate out of the screen).

jokasimr:
**Examples**

Screenshot from 2026-02-11 09-27-10
Screenshot from 2026-02-11 09-26-56

SimonHeybrock:
> Finally, I wonder if placing things at scale in the graph would be useful/possible for the translations? You could then build some sort of instrument layout with relative spacings between components?

I think that would make it unusable in many cases. Let us not do that

SimonHeybrock:
> I added fields for the transformation vector and the magnitude.

Can you also add add the others? No reason to omit anything, right?

src/scippnexus/nxtransformations.py
```diff
@@ -260,11 +260,113 @@ class TransformationChain(DependsOn):
...
+                                f'{value.unit}'
+                            )
+                        elif 'time' in value.sizes:
+                            vals = np.asarray(value.values, dtype=float)
+                            unit = value.unit
+                            label = (
+                                f'{label}\\nvalue ∈ '
+                                f'[{_fmt(np.nanmin(vals))} {unit}, '
+                                f'{_fmt(np.nanmax(vals))} {unit}]'
+                            )
```
SimonHeybrock: Why all the duplication and why not use Scipp?

src/scippnexus/nxtransformations.py
```diff
@@ -260,11 +260,113 @@ class TransformationChain(DependsOn):
...
+
+        depends_on = self
+        prev = None
+        visited = []
+        try:
+            while (path := depends_on.absolute_path()) is not None:
+                if path in visited:
+                    raise ValueError(
+                        f'Circular depends_on chain detected: {[*visited, path]}'
+                    )
```
SimonHeybrock: Why not visualize this? The point of the visualization is to debug the chain. I don't see the need to raise here?

src/scippnexus/nxtransformations.py
```diff
@@ -260,11 +260,113 @@ class TransformationChain(DependsOn):
...
+                dot.node(path, label=label, **attrs)
+                if prev is not None:
+                    dot.edge(prev, path, label='depends_on')
+                prev = path
+                depends_on = transform.depends_on
+        except KeyError as e:
+            m = f'depends_on chain {depends_on} references missing node {e}'.replace(
+                '\n', ''
+            )
+            warnings.warn(UserWarning(m), stacklevel=2)
```
SimonHeybrock: Why is this not included in the graph? Also not sure why we would want to warn here? Can that just be a warning color on the graph?

src/scippnexus/nxtransformations.py
jokasimr: Fixed now 👍

jokasimr:
>   Can you also add add the others? No reason to omit anything, right?

Including too much information in the graphviz boxes will make the visualization hard to read and understand. What fields are you thinking about specifically?

jokasimr:
Screenshot from 2026-02-11 16-29-03
Screenshot from 2026-02-11 16-28-53

src/scippnexus/nxtransformations.py
```diff
@@ -260,11 +260,105 @@ class TransformationChain(DependsOn):
...
+        except Exception as e:  # pragma: no cover - optional dependency
+            raise ImportError(
+                "The 'graphviz' package is required to visualize transformation chains."
+            ) from e
+
+        dot = Digraph()
+        dot.attr(rankdir=rankdir)
+
+        def _fmt(x: sc.DataArray | sc.Variable) -> str:
+            return f'{x.value:.3g} {x.unit}'
```
jl-wynen:
```suggestion
            return f'{x:c}'
```

src/scippnexus/nxtransformations.py
```diff
@@ -260,11 +260,105 @@ class TransformationChain(DependsOn):
...
+                    if ttype == 'translation':
+                        attrs.update({'shape': 'box'})
+                    else:
+                        attrs.update({'shape': 'ellipse'})
+                dot.node(path, label=label, **attrs)
+                if prev is not None:
+                    dot.edge(prev, path, label='depends_on')
+                prev = path
+                depends_on = transform.depends_on
+        except KeyError:
```
jl-wynen: This code is very long and deeply nested. And this try-except encompasses almost all of it. So it's difficult to understand how this works and what failures are supposed to trigger this fallback. Can you split this into smaller functions?

jokasimr: I've moved the try-except to only cover the part of the code where it's relevant.
This makes the code less nested and easier to read.
I don't think it would benefit from being split into smaller functions at this point.

SimonHeybrock:
> > Can you also add add the others? No reason to omit anything, right?
> 
> Including too much information in the graphviz boxes will make the visualization hard to read and understand. What fields are you thinking about specifically?

All the values/attrs that are part of a transform. If you omit some it may be misinterpreted as a bug, and not seeing all of them means it is impossible to interpret what the transform does. There are not many, so I think including all should be perfectly feasible?

SimonHeybrock (commented):
Thanks, code looks good. Can you share some screenshots of the latest state?

@jl-wynen Did you have more comments?

jokasimr:
Screenshot from 2026-03-02 07-48-48
Screenshot from 2026-03-02 07-47-58
Screenshot from 2026-03-02 07-47-36

SimonHeybrock (approved):

jl-wynen (approved):

PR #299, opened 2026-05-01
Allow empty slice when loading log with sublogs

jl-wynen:
Loading an NXlog that contians sublogs with `x[:]` raises an exception because we don't support slicing sublogs. But in this case, it should behave the same as `x[()]` or `x[...]` for 1D data. So this PR handles it this way.

I ran into this in ESSreduce when passing the default `TimeInterval` (`slice(None, None, None)`) to a loader for an NXlog.

### DIFF
--- a/src/scippnexus/_common.py
+++ b/src/scippnexus/_common.py
@@ -79,9 +79,6 @@ def to_canonical_select(
         return {}
     if isinstance(select, tuple) and len(select) == 0:
         return {}
-    if select == slice(None, None, None):
-        check_1d()
-        return {}
     if isinstance(select, tuple) and isinstance(select[0], str):
         key, sel = select
         return {key: sel}
--- a/src/scippnexus/_common.py
+++ b/src/scippnexus/_common.py
@@ -79,6 +79,9 @@ def to_canonical_select(
         return {}
     if isinstance(select, tuple) and len(select) == 0:
         return {}
+    if select == slice(None, None, None):
+        check_1d()
+        return {}
     if isinstance(select, tuple) and isinstance(select[0], str):
         key, sel = select
         return {key: sel}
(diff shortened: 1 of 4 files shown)


src/scippnexus/_common.py
```diff
@@ -79,6 +79,9 @@ def check_1d() -> None:
         return {}
     if isinstance(select, tuple) and len(select) == 0:
         return {}
+    if select == slice(None, None, None):
+        check_1d()
```
SimonHeybrock: Why is the branch below that does `elif isinstance(select, int | sc.Variable) or isinstance(select, slice):` not sufficient?

jl-wynen: Because that returns a 1d slice which leads to this error downstream:
```
     59 def read_children(self, sel: ScippIndex) -> sc.DataGroup:
     60     # Sublogs have distinct time axes (with a different length). Must disable
     61     # positional indexing.
     62     if self._sublogs and ('time' in to_canonical_select(list(self.sizes), sel)):
---> 63         raise sc.DimensionError(
     64             "Cannot positionally select time since there are multiple "
     65             "time fields. Label-based selection is not supported yet."
     66         )
     67     dg = super().read_children(sel)
     68     for name, field in self._sublog_children.items():

DimensionError: Cannot positionally select time since there are multiple time fields. Label-based selection is not supported yet.
```
https://github.com/scipp/scippnexus/blob/1ca907d98df2c9a94880760df089425e4f8d988d/src/scippnexus/nxlog.py#L59-L70

Alternatively, we could change the check here.

SimonHeybrock: Right, so this specifically fixes a bug when the ESS-style "sublogs" are present (which are not in the standard, iirc)? A more targeted fix might indeed make sense, but I don't mind too much.

Can you update the PR title/description?

I think this also means that the new tests do not really do much that didn't work before, as they contain no sublogs as far as I can see. The removed line in `test_log_with_connection_status_raises_with_positional_and_label_indexing` is I think the test you actually need, except that it would now *not* raise.

jl-wynen: Updated

SimonHeybrock (approved):

